PR #355, opened 2014-04-17
Fix linking problem on OS X for iconv

gravityrail:
On OS X Mavericks, the built-in version of iconv is missing the symbol "libiconv_open", resulting in an error like this when you build and test:

```
dyld: lazy symbol binding failed: Symbol not found: _libiconv_open
  Referenced from: /Users/dan/workspace/3dna/rugged/lib/rugged/rugged.bundle
  Expected in: flat namespace

dyld: Symbol not found: _libiconv_open
  Referenced from: /Users/dan/workspace/3dna/rugged/lib/rugged/rugged.bundle
  Expected in: flat namespace

rake aborted!
SIGTRAP
```

The Homebrew version of iconv is compatible, but in order to use that library we need to place system-level frameworks last in the cmake search path. That's what this patch does. I haven't had a chance to test on Linux or other systems so I don't know yet whether it breaks the package there.

### DIFF
--- a/ext/rugged/extconf.rb
+++ b/ext/rugged/extconf.rb
@@ -34,7 +34,8 @@ Dir.chdir(LIBGIT2_DIR) do
   Dir.mkdir("build") if !Dir.exists?("build")
 
   Dir.chdir("build") do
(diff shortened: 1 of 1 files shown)


volkanunsal:
Six years on, I'm having this issue on a Mac. I was able to get past it by adding this flag via ENV vars. I'd encourage you to make this flag `-DCMAKE_FIND_FRAMEWORK=LAST` default.

PR #420, opened 2014-09-14
Expose core dump on index.diff from merge.

cirosantilli:
I was trying to get the conflicts, when I tried this.

Now I think this should not do what I want: in theory the diff should be empty.

But I think it is not acceptable to get core dumps on Ruby library? Exceptions should be thrown instead?

Let's see if Travis reproduces.


arthurschreiber:
Hey @cirosantilli!

Sorry it took me such a long time to get back to you. I just verified that the bug you encountered here was coming from libgit2 and submitted a pull request to fix it over at https://github.com/libgit2/libgit2/issues/3173.

This only needs to be fixed in libgit2, no changes in rugged required.

cirosantilli:
Good to know! Cheers.

carlosmn:
Fixed via #574

PR #437, opened 2014-11-21
Implement Blob::Loader

vmg:
This will hopefully cut down on the amount of memory that Linguist uses when indexing repositories.


arthurschreiber:
@vmg Is there still a need for something like this? There's a way to make this work even with newer Ruby versions and the string sharing stuff.

vmg:
No, we've decided not to take this approach. :)

PR #453, opened 2015-01-29
Add tree_at similar to blob_at

sarupbanskota:
I was wondering if a `tree_at` method would be useful (similar to the `blob_at`? (I'm only recently exploring rugged's source, so I could be wrong). It's not a lot of effort doing that otherwise, but a method for that could keep things easy. 

There isn't much error handling atm.

### DIFF
--- a/lib/rugged/repository.rb
+++ b/lib/rugged/repository.rb
@@ -210,6 +210,17 @@ module Rugged
       remote_or_url.fetch(*args)
     end
 
+    # Get the tree at a path for a specific revision.
+    #
+    # revision - The String SHA1.
+    # path     - The String file path.
+    #
+    # Returns a String.
+    def tree_at(revision, path)
+      tree = Rugged::Commit.lookup(self, revision).tree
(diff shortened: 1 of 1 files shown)


lib/rugged/repository.rb
```diff
@@ -210,6 +210,17 @@ def fetch(remote_or_url, *args)
...
 
+    # Get the tree at a path for a specific revision.
+    #
+    # revision - The String SHA1.
+    # path     - The String file path.
+    #
+    # Returns a String.
+    def tree_at(revision, path)
+      tree = Rugged::Commit.lookup(self, revision).tree
+      tree.path(path)[:oid]
```
arthurschreiber: The way this is implemented right now, `#tree_at` would return the tree's oid, not the actual `Rugged::Tree` object. This is different from how `#blob_at` works.

Would you mind fixing this up and adding some test cases?

scutuatua-crypto (approved):

PR #456, opened 2015-02-03
fix SEGV in rugged_branch_collection.c

aiionx:
Check if git_branch_iterator_new returned error when initializing iter

Error might be returned if for some reason the iter could not be initialized,
but currently that error is not being checked.


aiionx:
Strange it fails, it just checks if error occurred.

vmg:
@aiionx: looks like you moved the file around in your working directory. The source needs to be under `ext/rugged` for it to compile. Can you please move it back? :)

aiionx:
Oh i didnt noticed. How strange. Ill close this and create a new pull request.

Thanks

aiionx:
https://github.com/libgit2/rugged/pull/457

PR #457, opened 2015-02-04
Fix SIGSEGV: check if there was an exception when initializing iterator

aiionx:
Currently there is no check if an exception was raised when initializing the iterator at each_branch function.


aiionx:
Not checking can produce a segmentation fault when trying to iterate over an uninitialized iterator.

vmg:
This looks right this time. :))

Thanks for the PR!

sandish2222:
:+1:

Issue #438, opened 2014-11-24
On OS X, '~' is not expanded properly


ysgard:
irb(main):055:0> Rugged::Repository.new('~/P4GitLab/p4-gitlab')
=> #<Rugged::Repository:70149931928520 {path: "/Users/juytven/P4GitLab/p4-gitlab/~/P4GitLab/p4-gitlab/.git/"}>
irb(main):056:0> Rugged::Repository.new('~/my-repo')
Rugged::OSError: Failed to resolve path '~/my-repo': No such file or directory
    from (irb):56:in `new'
    from (irb):56
    from /Users/juytven/.rbenv/versions/2.1.3/bin/irb:11:in`<main>'
irb(main):057:0> quit

➜  p4-gitlab git:(master) gem list rugged

**\* LOCAL GEMS ***

rugged (0.21.2, 0.21.0)
➜  p4-gitlab git:(master) ruby -v
ruby 2.1.3p242 (2014-09-19 revision 47630) [x86_64-darwin14.0]
➜  p4-gitlab git:(master)

carlosmn:
It looks like we forgot to close this issue. Indeed the tilde should be expanded by the method that does that, we take the path as-is.

PR #459, opened 2015-02-05
Remove the tilde from the Repository documentation

carlosmn:
Using the tilde in the documentation for repository initialization
suggests that we perform tilde expansion in the path provided. We do no
such thing, so remove it as it's quite unlikely someone does want a
directory called '~' to have their git repos.

---

Maybe we want to have `File.expand_path('~/test/.git')` instead? The current way should go either way, this is likely the cause for #438.


arthurschreiber:
:heart:

PR #464, opened 2015-03-05
Add Tree#count_recursive

adelcambre:
This adds a new `count_recursive` method to Rugged tree objects. This counts all blobs in the tree recursively, with an optional limit to bail early. This allows asking things like "Are there more than 1 million files in this repo?" without a roundtrip into ruby for each file.

### DIFF
--- a/ext/rugged/rugged.h
+++ b/ext/rugged/rugged.h
@@ -149,8 +149,8 @@ struct rugged_cb_payload
 
 struct rugged_treecount_cb_payload
 {
-	unsigned int count;
-	unsigned int limit;
+	int count;
+	int limit;
 };
 
 struct rugged_remote_cb_payload
--- a/ext/rugged/rugged.h
+++ b/ext/rugged/rugged.h
@@ -150,7 +150,7 @@ struct rugged_cb_payload
 struct rugged_treecount_cb_payload
 {
 	unsigned int count;
-	int limit;
+	unsigned int limit;
 };
 
 struct rugged_remote_cb_payload
--- a/ext/rugged/rugged.h
+++ b/ext/rugged/rugged.h
@@ -147,12 +147,6 @@ struct rugged_cb_payload
     int exception;
 };
 
-struct rugged_treecount_cb_payload
-{
-	int count;
-	int limit;
-};
-
 struct rugged_remote_cb_payload
 {
 	VALUE progress;
--- a/ext/rugged/rugged.h
+++ b/ext/rugged/rugged.h
@@ -147,6 +147,12 @@ struct rugged_cb_payload
     int exception;
 };
 
+struct rugged_treecount_cb_payload
+{
+	unsigned int count;
+	int limit;
+};
+
 struct rugged_remote_cb_payload
 {
 	VALUE progress;
--- a/ext/rugged/rugged_tree.c
+++ b/ext/rugged/rugged_tree.c
@@ -96,7 +96,6 @@ static VALUE rb_git_tree_entrycount(VALUE self)
 static int rugged__treecount_cb(const char *root, const git_tree_entry *entry, void *payload)
 {
 	int *count = (int *)payload;
-  printf("%s: %d\n", git_tree_entry_name(entry), git_tree_entry_type(entry));
 
   if(git_tree_entry_type(entry) == GIT_OBJ_TREE) {
     return 0;
--- a/ext/rugged/rugged_tree.c
+++ b/ext/rugged/rugged_tree.c
@@ -109,10 +109,14 @@ static int rugged__treecount_cb(const char *root, const git_tree_entry *entry, v
 
 /*
  *  call-seq:
- *    tree.count_recursive -> count
- *    tree.length_recursive -> count
+ *    tree.count_recursive(limit=nil) -> count
  *
- *  Return the number of blobs contained in the tree and all subtrees.
+ *  `limit` - The maximum number of blobs to the count in the repository.
+ *  Rugged will stop walking the tree after `limit` items to avoid long
+ *  execution times.
+ *
+ *  Return the number of blobs (up to the limit) contained in the tree and
+ *  all subtrees.
  */
 static VALUE rb_git_tree_entrycount_recursive(int argc, VALUE* argv, VALUE self)
 {
--- a/ext/rugged/rugged_tree.c
+++ b/ext/rugged/rugged_tree.c
@@ -124,17 +124,17 @@ static VALUE rb_git_tree_entrycount_recursive(int argc, VALUE* argv, VALUE self)
 	Data_Get_Struct(self, git_tree, tree);
 	int error;
 	struct rugged_treecount_cb_payload payload;
-	VALUE rbLimit;
+	VALUE rb_limit;
 
-	if (rb_scan_args(argc, argv, "01", &rbLimit) == 0)
-		rbLimit = Qnil;
+	if (rb_scan_args(argc, argv, "01", &rb_limit) == 0)
+		rb_limit = Qnil;
 
 	payload.limit = -1;
 	payload.count = 0;
 
-	if (!NIL_P(rbLimit)) {
-		Check_Type(rbLimit, T_FIXNUM);
-		payload.limit = FIX2INT(rbLimit);
+	if (!NIL_P(rb_limit)) {
+		Check_Type(rb_limit, T_FIXNUM);
+		payload.limit = FIX2INT(rb_limit);
 	}
 
 
--- a/ext/rugged/rugged_tree.c
+++ b/ext/rugged/rugged_tree.c
@@ -95,7 +95,7 @@ static VALUE rb_git_tree_entrycount(VALUE self)
 
 static int rugged__treecount_cb(const char *root, const git_tree_entry *entry, void *payload)
 {
-	unsigned int *count = (int *)payload;
+	unsigned int *count = (unsigned int *)payload;
 
 	if(git_tree_entry_type(entry) == GIT_OBJ_TREE) {
 		return 0;
--- a/ext/rugged/rugged_tree.c
+++ b/ext/rugged/rugged_tree.c
@@ -93,6 +93,40 @@ static VALUE rb_git_tree_entrycount(VALUE self)
 	return INT2FIX(git_tree_entrycount(tree));
 }
 
+static int rugged__treecount_cb(const char *root, const git_tree_entry *entry, void *payload)
+{
+	int *count = (int *)payload;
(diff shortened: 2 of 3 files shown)


ext/rugged/rugged.h
```diff
@@ -147,6 +147,12 @@ struct rugged_cb_payload
     int exception;
 };
 
+struct rugged_treecount_cb_payload
+{
+	unsigned int count;
+	int limit;
```
brianmario: `unsigned` here as well?

adelcambre: Oh yeah, need to do that.

ext/rugged/rugged_tree.c
```diff
@@ -93,6 +93,63 @@ static VALUE rb_git_tree_entrycount(VALUE self)
...
+
+	if (rb_scan_args(argc, argv, "01", &rbLimit) == 0)
+		rbLimit = Qnil;
+
+	if (NIL_P(rbLimit))
+		payload.limit = -1;
+	else if (TYPE(rbLimit) != T_FIXNUM)
+		rb_raise(rb_eTypeError, "limit must be a Fixnum");
+	else
+		payload.limit = FIX2INT(rbLimit);
```
brianmario: this could maybe be simplified to something like:

``` C
if (!NIL_P(rbLimit)) {
  Check_Type(rbLimit, T_FIXNUM);
  payload.limit = FIX2INT(rbLimit);
}
```

`Check_Type` will raise an `ArgumentError` for you.

ext/rugged/rugged.h
```diff
@@ -147,6 +147,12 @@ struct rugged_cb_payload
     int exception;
 };
 
+struct rugged_treecount_cb_payload
+{
+	unsigned int count;
+	unsigned int limit;
```
arrbee: You're using `-1` as a "no limit" signifier below, so either this has to go back to a signed type (because otherwise `payload->limit >= 0` will always be true) or you can special case `0` instead.

Personally, I would probably make these both `ssize_t` values. If `ssize_t` isn't easily usable from within Rugged for portability reasons, you could make both simple `int` values, too.

adelcambre: Totally correct, was just chatting about that with @brianmario. I think just `int` is probably fine. What are the portability concerns with `ssize_t`?

ext/rugged/rugged_tree.c
```diff
@@ -93,6 +93,63 @@ static VALUE rb_git_tree_entrycount(VALUE self)
...
+
+	if (payload->limit >= 0 && payload->count >= payload->limit)
+		return -1;
+	else if(git_tree_entry_type(entry) == GIT_OBJ_TREE)
+		return 0;
+	else {
+		++(payload->count);
+		return 1;
+	}
+}
```
arrbee: Unfortunately in the libgit2 `git_tree_walk` API, there is no maximum depth check and the implementation is recursive, so there are some theoretical DOS vectors here. I think probably the best bet is to fix it at the libgit2 level and either reimplement `git_tree_walk` using the `git_iterator_for_tree` API or simply add a depth check internal to the API. (It would be nice to reflect the depth as a parameter to the callback, but that would be a breaking API change, so probably not worth it.)

ext/rugged/rugged_tree.c
adelcambre: Oh cool. I'll switch it to this. Didn't know about `Check_Type`.

ext/rugged/rugged.h
arrbee: I just know in libgit2 we used to avoid it in public headers because they way you include the typedef for it can vary from platform to platform. (Interestingly, it looks like it had recently crept in by way of the new `include/git2/sys/stream.h` header.)

ext/rugged/rugged_tree.c
adelcambre: Hmm, good point. Rather than just counting blob, I could count all tree entries (including subtrees). This means that providing a limit will give protection from at least one more DOS vector. Would that cover the case you're thinking of?

arrbee: Regrettably, I think not. The issue is actually hard to compensate for except at the libgit2 level (without introducing a waste of time scanning path data) - namely that you can create a 100,000 level deep tree and not hit your walk limit, but burn up a huge amount of stack space internal to libgit2. Let me look over there a bit...

ext/rugged/rugged_tree.c
```diff
@@ -93,6 +93,63 @@ static VALUE rb_git_tree_entrycount(VALUE self)
...
+ */
+static VALUE rb_git_tree_entrycount_recursive(int argc, VALUE* argv, VALUE self)
+{
+	git_tree *tree;
+	Data_Get_Struct(self, git_tree, tree);
+	int error;
+	struct rugged_treecount_cb_payload payload;
+	VALUE rbLimit;
+
+	if (rb_scan_args(argc, argv, "01", &rbLimit) == 0)
```
arthurschreiber: You can remove the `if` condition and change this to just:

`rb_scan_args(argc, argv, "01", &rbLimit);`

`rb_scan_args` will set `rbLimit` to nil if it was not passed.

ext/rugged/rugged_tree.c
vmg: fwiw MRI cannot stack overflow because it protects the end of the stack in the VM and will gracefully raise a `StackOverflow` exception into Ruby-land if a C extension overflows... So I'm not sure I'd consider this an attack vector.

ext/rugged/rugged_tree.c
vmg: We only use caps for class and module names, so `rbLimit` should be `rb_limit` here.

adelcambre: @arthurschreiber Coulda sworn I tried that, but that's way better. I'll switch to that. 

@vmg Still getting used to the C-ext conventions. Thanks.

ext/rugged/rugged_tree.c
arrbee: Ah, thank you @vmg for bringing actual knowledge to the table. :kissing_cat:

ext/rugged/rugged_tree.c
```diff
@@ -93,6 +93,62 @@ static VALUE rb_git_tree_entrycount(VALUE self)
...
+
+	if (!NIL_P(rb_limit)) {
+		Check_Type(rb_limit, T_FIXNUM);
+		payload.limit = FIX2INT(rb_limit);
+	}
+
+
+	error = git_tree_walk(tree, GIT_TREEWALK_PRE, &rugged__treecount_cb, (void *)&payload);
+
+	if(error && giterr_last()->klass == GITERR_CALLBACK) {
```
vmg: Neat. :+1:

ext/rugged/rugged.h
```diff
@@ -147,6 +147,12 @@ struct rugged_cb_payload
     int exception;
 };
 
+struct rugged_treecount_cb_payload
```
vmg: This doesn't need to be in the external header because it's only used for the recursive count. Can you move it next to the function? That'll make the implementation easier to follow.

(the other payloads are here because they are used by different files)

adelcambre: Absolutely. I actually moved it in here as that's where it looked like all of the other payloads were defined. I'll move it back.

vmg:
Neat stuff. Props for your first Rugged PR!

adelcambre:
> first Rugged PR

Heh! I think this is my 3rd or 4th!

adelcambre:
Anything else need done here? Am I good to go ahead and merge myself?

ext/rugged/rugged_tree.c
```diff
@@ -93,6 +93,68 @@ static VALUE rb_git_tree_entrycount(VALUE self)
...
+ *  Rugged will stop walking the tree after `limit` items to avoid long
+ *  execution times.
+ *
+ *  Return the number of blobs (up to the limit) contained in the tree and
+ *  all subtrees.
+ */
+static VALUE rb_git_tree_entrycount_recursive(int argc, VALUE* argv, VALUE self)
+{
+	git_tree *tree;
+	Data_Get_Struct(self, git_tree, tree);
```
arthurschreiber: You have to move this statement below other variable declarations, as this generates a warning. (See the travis build logs).

ext/rugged/rugged_tree.c
arthurschreiber: Nitpick: Add a whitespace after `if`.

ext/rugged/rugged_tree.c
```diff
@@ -93,6 +93,68 @@ static VALUE rb_git_tree_entrycount(VALUE self)
...
+{
+	int count;
+	int limit;
+};
+
+static int rugged__treecount_cb(const char *root, const git_tree_entry *entry, void *data)
+{
+	struct rugged_treecount_cb_payload *payload = data;
+
+	if (payload->limit >= 0 && payload->count >= payload->limit)
```
arthurschreiber: Can you add braces here? I know we generally don't use braces for single-line bodies, but I feel this looks a bit weird because of the last else condition which spans multiple lines.

ext/rugged/rugged_tree.c
```diff
@@ -93,6 +93,68 @@ static VALUE rb_git_tree_entrycount(VALUE self)
...
+		return -1;
+	else if(git_tree_entry_type(entry) == GIT_OBJ_TREE)
```
arthurschreiber: Nitpick: Add a whitespace after `if`.

arthurschreiber:
Except for the last few nitpicks, this looks great! :+1:

adelcambre:
@arthurschreiber Cool, thanks for the feedback! Got those all updated.

vmg:
:sparkles:

adelcambre:
@vmg :heart_eyes_cat: Do you mind pushing up a new beta version of the gem so I can get it merged on our end?

vmg:
@adelcambre: For the betas we usualy just SHA-vendor them in GitRPC instead of pushing to Rubygems

arthurschreiber:
I really need to kick out a new "stable" release with libgit2 updated to 0.22 so we can start development on 0.23... :disappointed:

vmg:
@arthurschreiber: can you start by pushing a new beta to Rubygems with these changes? Looks like @adelcambre needs a release to vendor in GitRPC.

arthurschreiber:
Sure, I'll cut a release later today.

arthurschreiber:
Yo @adelcambre - I just released `0.22.1b1`. Hope that helps. :sparkles:

PR #468, opened 2015-03-19
Bump libgit2 to 8311db0cf8ae15b46edd14358a8238862e0bac4d

simonsj:
Rev to the latest libgit2.

My motivation here is to pull in [this fix](https://github.com/libgit2/libgit2/pull/3000) though I don't know what other changes are included.

cc @vmg @arthurschreiber

PR #470, opened 2015-03-19
Update to latest libgit2

arthurschreiber:
This updates to the latest libgit2 version. There is a single test failure, but that will be fixed once #469 gets merged.

This closes #468.

//cc @vmg @simonsj

### DIFF
--- a/ext/rugged/rugged_branch_collection.c
+++ b/ext/rugged/rugged_branch_collection.c
@@ -120,12 +120,6 @@ static git_branch_t parse_branch_type(VALUE rb_filter)
  *    Overwrites the branch with the given +name+, if it already exists,
  *    instead of raising an exception.
  *
- *  :message ::
- *    A single line log message to be appended to the reflog.
- *
- *  :signature ::
- *    The signature to be used for populating the reflog entry.
- *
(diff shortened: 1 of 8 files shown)


vmg:
Looking very nice. Thanks for rushing this @arthurschreiber!

Random question, now that are these signature APIs are gone, how do you pass signatures around? Are there other APIs we haven't wrapped yet that do this? cc @carlosmn

carlosmn:
The default signature will be taken from the environment, just like git does. If you want to override the signature, you can call `git_repository_set_iden()` and we'll use that.

arthurschreiber:
I don't think this function is bound yet. I'll get to it later today.

vmg:
@arthurschreiber: Can you fix the test and then go ahead and merge? Thanks!

arthurschreiber:
I just rebased this against the latest master, which should fix the test failure.

arthurschreiber:
:heart:

vmg:
:+1::+1:

adelcambre:
Thanks @arthurschreiber! Can you release a new beta on rubygems? Thanks!

PR #483, opened 2015-05-07
Memory-safe Walker API

vmg:
Here's a new APi for `Walker` which is memory safe (i.e. frees the walker immediately and prevents used packfiles from lingering around). The walker code has been refactored to reduce duplication; old APIs should be 100% backwards-compatible.

@arthurschreiber can you fit this on the release? Thanks :)

cc @spraints who is already using this on production

### DIFF
--- a/ext/rugged/rugged_revwalk.c
+++ b/ext/rugged/rugged_revwalk.c
@@ -302,13 +302,6 @@ static VALUE do_walk(VALUE _payload)
 	return Qnil;
 }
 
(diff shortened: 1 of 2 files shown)


spraints:
:+1:

arthurschreiber:
This is looking great! I'll cut a new release containing this and the libgit2 bump asap.

arthurschreiber:
@vmg This was released as `rugged@0.23.0b2`.

PR #486, opened 2015-05-17
Update to the latest libgit2 changes

arthurschreiber:
This mainly contains changes necessary to support the API changes in the `git_remote_*` API that were recently introduced.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ matrix:
   fast_finish: true
   allow_failures:
     - rvm: rbx-2
+    - rvm: ruby-head
 
 install:
   - ./vendor/libgit2/script/install-deps-${TRAVIS_OS_NAME}.sh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@
 
     *Arthur Schreiber*
 
+*   `Rugged::Remote` instances are now immutable.
+
+    * `Remote#clear_refspecs` and `Remote#save` were removed without
+      replacement.
+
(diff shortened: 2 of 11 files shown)


carlosmn:
This may not be an important distinction in many cases, but remotes aren't technically immutable. They do have internal mutable state which is where we keep the data for fetches and pushes. Maybe a more accurate way of describing it would be that you cannot change their properties/attributes?

arthurschreiber:
@carlosmn True. You can still modify them in whatever way you like (especially thanks to ruby being a very dynamic language). I'll reword this.

arthurschreiber:
@vmg Thanks! Do you want me to fix this up, merge and cut a new release later today?

vmg:
@arthurschreiber: There is a regression on the diff code, we're trying to figure this out.

arthurschreiber:
@vmg I'll pull in the latest libgit2, fix everything up and release a new version this evening.

vmg:
Cool. See if you can work around the submodule changes and release the version. Thanks!

arthurschreiber:
@vmg I updated the code to work with the submodule changes.

https://github.com/libgit2/libgit2/pull/3258 needs to be merged and the submodule API in Rugged needs to be changed a bit to make sense again, but then we're good for a new release.

vmg:
@arthurschreiber: Any more changes related to submodules you'd like to do? I'm blocked on this PR.

arthurschreiber:
Yeah, right now, the API makes no sense. E.g. if you change a submodule option like the ignore behaviour, that change gets persisted to disk, but the live object you get does not get updated. If that is no issue for you, we can merge this and I can fix it later today.

PR #491, opened 2015-06-24
Implement Blob#loc

vmg:
Pretty straightforward. Allows us to count lines on a blob without getting it into Rubyland.

cc @aroben

### DIFF
--- a/ext/rugged/rugged_blob.c
+++ b/ext/rugged/rugged_blob.c
@@ -317,6 +317,44 @@ static VALUE rb_git_blob_from_io(int argc, VALUE *argv, VALUE klass)
 	return rugged_create_oid(&oid);
 }
 
+/*
+ *  call-seq:
+ *    blob.loc -> int
+ *
+ *  Return the number of lines for this blob,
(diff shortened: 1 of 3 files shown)


arthurschreiber:
LGTM. :+1:

ext/rugged/rugged_blob.c
```diff
@@ -317,6 +317,44 @@ static VALUE rb_git_blob_from_io(int argc, VALUE *argv, VALUE klass)
...
+	data_end = data + git_blob_rawsize(blob);
+
+	if (data == data_end)
+		return INT2FIX(0);
+
+	for (; data < data_end; ++data) {
+		if (data[0] == '\n') {
+			loc++;
+		}
+		else if (data[0] == '\r') {
```
arthurschreiber: Maybe I'm missing something, but is there any need for this else condition?

arthurschreiber: Ah, crap. Ignore my comment. :dizzy_face:

aroben:
Thanks!

PR #492, opened 2015-06-26
Change the Submodule API.

arthurschreiber:
This PR aligns the Submodule API with the new libgit2 submodule API.

I pulled out all setter methods that were available on `Rugged::Submodule` into a common `#update` method on the `SubmoduleCollection` class.

### DIFF
--- a/ext/rugged/rugged_submodule.c
+++ b/ext/rugged/rugged_submodule.c
@@ -559,34 +559,6 @@ static VALUE rb_git_submodule_url(VALUE self)
 
 /*
  *  call-seq:
- *    submodule.url = url -> url
- *
- *  Set the URL in the config for the submodule.
- *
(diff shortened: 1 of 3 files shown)


arthurschreiber:
Here's a bit of background on why I made this change in the first place:

@carlosmn removed the pseudo ORM for submodule in libgit2 itself, and after updating to the new function signatures, the existing submodule API got quite confusing: Calling a setter method on a submodule instance would update the on-disk configuration, but not update the submodule instance itself. The user would have to call `#reload` manually to load back these changes into the instance.

Changing submodule instances to be unchangable after they were loaded makes this issue go away and aligns submodules with other rugged objects (like references, tags, etc).

vmg:
Looks great :D

arthurschreiber:
I'll probably prepare something similar for the `Remote` API as well.

PR #494, opened 2015-06-29
Rugged 0.23b4

vmg:


arthurschreiber:
LGTM. Can you also make sure you bump the version number in the `Gemfile.lock`?

vmg:
I'm not sure of why the `Gemfile.lock` was checked in, but that's a terrible practice for gems. I've removed it and stopped travis from doing deployment installs (??)

arthurschreiber:
@vmg The `Gemfile.lock` is never released with the gem, and I'm pretty sure it should be in the repo.

It is only used when running rake and other stuff using `bundle exec ...`. It ensures tests run with the correct **development dependencies**. `Gemfile` and `Gemfile.lock` don't have any impact on users of the rugged gem.

arthurschreiber:
After some clarification, I'm :+1: with dropping `Gemfile.lock`.

It did not have any impact on dependency resolution when installing/using rugged from another application, but dropping it will allow us to identify potential incompatibilities with new versions of rugged's dependencies earlier and we won't have to update the `Gemfile.lock` after each version bump anymore.

PR #497, opened 2015-07-03
Better Fixture Management

arthurschreiber:
Fixture Management in Rugged has gone a bit out of hand. There's multiple different ways how fixtures are handled and managed, and it's not obvious when to use what.

* `Rugged::RepositoryAccess`: Including this module opens the bare repository at `test/fixtures/testrepo.git` and assigns it to `@repo`. Any modification to `@repo` will cause other test cases to fail, so this is only used for test cases that are read-only.
* `Rugged::TempRepositoryAccess`: This creates a clone of the repository at `test/fixtures/testrepo.git` and assigns it to `@repo`. This is used by test cases that modify the repo or that need a repo with a workdir.
* `Rugged::SandboxedTestCase`: This is a test class that provides the methods `sandbox_init` and `sandbox_clone`. `sandbox_init` copies test fixtures from libgit2 and returns a `Repository` for this copy, while `sandbox_clone` creates a clone for a repository that was previously copied via `sandbox_init`.
* `Rugged::SubmoduleTestCase`: This inherits from `Rugged::SandboxedTestCase` and provides another method to initialize a repository containing submodules.

---

This Pull Request tries to normalize and simplify the fixture repo handling. It replaces all the existing methods of creating test repositories with a single module containing a bunch of module methods:

* `FixtureRepo.empty`: Returns a new, empty repository.
* `FixtureRepo.from_rugged`: Returns a copy of a repository from the set of fixtures included in rugged.
* `FixtureRepo.from_libgit2`: Returns a copy of a repository from the set of fixtures included in libgit2.
* `FixtureRepo.clone`: Returns a clone of a `Rugged::Repository` instance.

All these methods operate on copies that are stored in temporary directories and that are cleaned up automatically through finalizers on the respective `Rugged::Repository` instances. This greatly simplifies test setup and teardown basically happens in a fully automated fashion.

---

TODO:

* [x] Fold `Rugged::SubmoduleTestCase` functionality into the methods provided by `FixtureRepo`.
* [x] ~~Check whether test cases fully pass on windows.~~ I couldn't get the latest rugged versions to compile and run on Windows.

### DIFF
--- a/rugged.gemspec
+++ b/rugged.gemspec
@@ -30,5 +30,5 @@ for testing and using the libgit2 library in a language that is awesome.
 desc
   s.add_development_dependency "rake-compiler", ">= 0.9.0"
   s.add_development_dependency "pry"
-  s.add_development_dependency "minitest", "~> 3.0.0"
+  s.add_development_dependency "minitest", "~> 3.0"
 end
--- a/test/blame_test.rb
+++ b/test/blame_test.rb
@@ -1,16 +1,10 @@
 require "test_helper"
 require "time"
 
-class BlameTest < Rugged::SandboxedTestCase
+class BlameTest < Rugged::TestCase
   def setup
-    super
-    @repo = sandbox_init("testrepo")
+    @repo = FixtureRepo.from_libgit2("testrepo")
     @blame = Rugged::Blame.new(@repo, "branch_file.txt")
-  end
-
-  def teardown
-    @repo.close
-    super
   end
 
   def test_blame_index
--- a/test/blob_test.rb
+++ b/test/blob_test.rb
@@ -1,7 +1,9 @@
 require "test_helper"
 
 class BlobTest < Rugged::TestCase
-  include Rugged::RepositoryAccess
+  def setup
+    @repo = FixtureRepo.from_rugged("testrepo.git")
+  end
 
   def test_lookup_raises_error_if_object_type_does_not_match
     assert_raises Rugged::InvalidError do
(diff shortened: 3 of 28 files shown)


vmg:
I had a chat with @charliesome about this. The new `FixtureRepo` design is overall great, and the test suite looks much cleaner, but now that we've gotten rid of the messy stuff (the module `include`s and all that), I see no reason why we shouldn't stick to an explicit `FixtureRepo#teardown`.

It makes the code a tad more verbose, but it has many advantages over the GC destructor approach. MRI's GC is particularly non-deterministic, so overall I'm not fond of the idea of depending on it for resource cleanups -- specially in cases where the test suite can crash. This can potentially leak a lot of data on disk.

arthurschreiber:
Thanks for having a look and thanks for the feedback! :)

> MRI's GC is particularly non-deterministic, so overall I'm not fond of the idea of depending on it for resource cleanups -- specially in cases where the test suite can crash. This can potentially leak a lot of data on disk.

Right. But if the tests crash, even in the case of an explicit `teardown` method, we'll still have leftover data on disk that won't be removed automatically. The only difference will be the amount of leftover data. I just checked and it looks like the amount of pending finalizer calls hover's around 10 (+/-5) during the test runs. But this might be very different depending on the Ruby version and other factors.

I'll see how much verbosity adding an explicit teardown method would add.

arthurschreiber:
Okay. So I changed things around a bit.

Fixtures are now automatically deleted after every test case, but a deletion via a finalizer is still scheduled in addition. (This is mainly done to support running rugged test cases on win32, which I wanted to be able to do for quite some time.)

arthurschreiber:
I think this is ready.

Cleanup of temp directories happens after each test case execution automatically, and does not rely on object finalizers.

vmg:
Congratulations on this PR, Arthur! It's everything I had in mind. :heart_eyes: 

Sure with the libgit2 suite looked as clean as this. :)

arthurschreiber:
:heart:

PR #500, opened 2015-07-15
Update Travis builds to use latest ruby versions

arthurschreiber:


arthurschreiber:
Looks like Travis does not support `2.1.6` yet on OS X: https://travis-ci.org/libgit2/rugged/jobs/71163494

PR #501, opened 2015-07-16
Add support for SSH credentials from memory.

arthurschreiber:
The build is currently failing because SSH memory credential support is not listed as part of `Rugged.features`.

@carlosmn Would a patch to libgit2 to add something like `GIT_FEATURE_SSH_KEY_FROM_MEMORY` be accepted?

### DIFF
--- a/ext/rugged/rugged_cred.c
+++ b/ext/rugged/rugged_cred.c
@@ -29,6 +29,7 @@ VALUE rb_mRuggedCred;
 VALUE rb_cRuggedCredUserPassword;
 VALUE rb_cRuggedCredSshKey;
 VALUE rb_cRuggedCredSshKeyFromAgent;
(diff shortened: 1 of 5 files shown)


carlosmn:
I was actually thinking about that recently. It would make sense to expose this as a feature.

arthurschreiber:
Okay. I'll see if I can hack this up, until then this PR will have to wait a bit. :smile:

mensfeld:
will it be merged? I owuld use this feature as well.

rxbynerd:
@arthurschreiber anything I can do to assist with getting this merged? happy to help contribute to rugged or libgit2 if needed

jpatters:
@carlosmn Any reason this was never added/addressed ?

scutuatua-crypto (approved):

Issue #441, opened 2014-12-01
Update Commit example documentation to mention index.write after committing


nanodeath:
The documentation mentions, in [one place](https://github.com/libgit2/rugged#writing-to-a-repository), at least, how to commit a newly written index (using `index.write_tree(repo)`), but fails to mention that index.write should be called after the commit has been performed.  This is a bit unintuitive to me, so it'd be nice if the API could be improved, but barring that, at least add that one line to the documentation (`index.write`).

---

This issue seems somewhat common -- it's come up at least twice: [on GitHub](https://github.com/libgit2/rugged/issues/254) as well as [on StackOverflow](https://stackoverflow.com/questions/24551308/how-can-i-use-rugged-to-create-and-commit-a-file-like-from-the-command-line?rq=1).  On the SO question, Carlos does say `That example does not emulate the git-commit command, it just creates a new commit.`, but since some people (at least me and two others) expect it to work this way, it seems like it'd be worth calling out, at least.

PR #502, opened 2015-07-17
Update README.md

saraid:
Noting that `index.write` is a necessary action after modifying the index.

Ref: #441

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ oid = repo.write("This is a blob.", :blob)
 index = repo.index
 index.read_tree(repo.head.target.tree)
 index.add(:path => "README.md", :oid => oid, :mode => 0100644)
+index.write
 
 options = {}
 options[:tree] = index.write_tree(repo)
@@ -377,6 +378,9 @@ index.add(ientry)
 
 # Stage. Create ientry from file in path, updates the index.
 index.add(path)
+
+# Write the index back to memory.
+index.write
 ```
 
 ---


README.md
```diff
@@ -377,6 +378,9 @@ index.add(ientry)
 
 # Stage. Create ientry from file in path, updates the index.
 index.add(path)
+
+# Write the index back to memory.
+index.write
```
carlosmn: This writes the index to disc.

README.md
```diff
@@ -152,6 +152,7 @@ oid = repo.write("This is a blob.", :blob)
 index = repo.index
 index.read_tree(repo.head.target.tree)
 index.add(:path => "README.md", :oid => oid, :mode => 0100644)
+index.write
```
carlosmn: This use-case does not want to persist the index, it's using it as a throw-away data structure (though it'd probably be good to replace `index = repo.index` to `index = Index.new`.

arthurschreiber:
I'm with @carlosmn here. The index actions that are shown in the Readme are not supposed to be written back to the repo index.

We probably still want to extend the documentation to cover `Index#write`, but not the way you've changed it here.

arthurschreiber:
Hey @saraid,

I'm going to close this, as the changes here don't actually reflect the different actions that are demonstrated in the Readme. Feel free to submit a different PR that changes the Readme to mention `Index#write`.

Thanks!

PR #509, opened 2015-08-11
Releases and pre-releases via Travis

arthurschreiber:
This enables automatic releases and pre-releases via Travis.

### Releases

Releases get cut when a new tag gets pushed to the repo and the build succeeds for the Ruby `2.2.2` build on travis. I'm thinking about automatically generating the `lib/rugged/version.rb` file based on the tag name, which would basically allow us to do release without having to bump this version identifier for each release.

### Pre-releases

Each commit that happens on the master branch will trigger the push of a pre-release gem to rubygems if the build on Ruby `2.2.2` passses. These gems will be versioned as `$version.$build_number.pre`. Using the build number is not perfect as it's not per-actual-version, but it's an increasing sequence, so if two pre-releases share the same "actual" release version, the one with the higher build number will be never, and gem version specifiers in `Gemfiles` or `*.gemspec` files will be handled correctly.

//cc @vmg You like?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,9 @@
-# Travis-CI Build for rugged
-# see travis-ci.org for details
-
 language: ruby
 
 os:
   - linux
   - osx
 
-# Ruby versions to test against
 rvm:
   - 1.9.3
   - 2.0.0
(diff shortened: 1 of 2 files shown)


arthurschreiber:
Instead of using the travis build number, we could also add something like this in the `travis.yml` file:

```yaml
before_deploy:
  - echo "module Rugged" > lib/rugged/version.rb
  - echo "  Version = VERSION = \"$(git describe --tags | tail -c +2)\"" >> lib/rugged/version.rb
  - echo "end"  >> lib/rugged/version.rb
```

vmg:
I really dig it. Let's use the `describe` output instead of the build number, because it's more stable and more commonly used.

arthurschreiber:
Okay, I changed this to get the tags from the `git describe` output. Example version number:

```
[rugged] main> Rugged::Version
=> "0.23.1.2.gdfb7659.pre"
```

arthurschreiber:
Ok, let's try this. :bomb:

PR #514, opened 2015-08-24
Add Rugged.valid_full_oid? helper method

brianmario:
Previously I've been using regex to check the validity of a string as a full 40-character sha1, but decided to benchmark that against the `Rugged.hex_to_raw` method and found it to be ~6x faster. Only bummer was that `hex_to_raw` will raise if the input string was invalid, and in that case ended up only being about 1.5x faster.

So this extracts just the validity check and returns a boolean value. Meaning we can skip the raise/rescue slowdown as well as creating the output string that `hex_to_raw` would have made.

### DIFF
--- a/ext/rugged/rugged.c
+++ b/ext/rugged/rugged.c
@@ -111,6 +111,28 @@ static VALUE rb_git_features(VALUE self)
 
 /*
  *  call-seq:
+ *    Rugged.valid_full_oid?(oid) -> true or false
+ *
+ *  Checks to see if a string contains a full 40-character sha1.
+ *
+ *    Rugged.valid_full_oid?('d8786bfc97485e8d7b19b21fb88c8ef1f199fc3f')
+ *    #=> true
+ */
+static VALUE rb_git_valid_full_oid(VALUE self, VALUE hex)
+{
+	git_oid oid;
+
+	Check_Type(hex, T_STRING);
+	int errorcode = git_oid_fromstr(&oid, StringValueCStr(hex));
+	if (errorcode < 0) {
+		return Qfalse;
+	} else {
+		return Qtrue;
+	}
+}
+
+/*
+ *  call-seq:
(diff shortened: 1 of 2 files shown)


ext/rugged/rugged.c
```diff
@@ -111,6 +111,28 @@ static VALUE rb_git_features(VALUE self)
...
+ *
+ *    Rugged.valid_full_oid?('d8786bfc97485e8d7b19b21fb88c8ef1f199fc3f')
+ *    #=> true
+ */
+static VALUE rb_git_valid_full_oid(VALUE self, VALUE hex)
+{
+	git_oid oid;
+
+	Check_Type(hex, T_STRING);
+	int errorcode = git_oid_fromstr(&oid, StringValueCStr(hex));
```
vmg: Declarations at the beginning of the block, please. :)

brianmario: oookaaayyy finneeee...

done :)

carlosmn: Those 25yo standards aren't going to adhere to themselves

brianmario: Doesn't C99 allow for them to be declared anywhere though?

vmg:
:))))

ext/rugged/rugged.c
ethomson: Yes, but Microsoft went through a long period of not giving a shit about C, and they didn't bother to include support for this until VS 2013.  Sorry!

brianmario: > Yes, but Microsoft went through a long period of not giving a shit about C, and they didn't bother to include support for this until VS 2013.

LOL that's right! I actually just went through that pain in another project. VS 2015 is actually WAY better in that regard.

Anyway, thanks for the review everyone! :)

carlosmn: C89 fo' lyfe, yo

aroben:
Nice!

PR #516, opened 2015-09-04
Add custom headers to smart http requests

spraints:
This integrates https://github.com/libgit2/libgit2/pull/3411 into rugged.

### DIFF
--- a/ext/rugged/rugged_remote.c
+++ b/ext/rugged/rugged_remote.c
@@ -181,6 +181,15 @@ void rugged_remote_init_callbacks_and_payload_from_options(
 	}
 }
 
+static void init_custom_headers(VALUE rb_options, git_strarray *custom_headers)
+{
+	if (!NIL_P(rb_options))
+	{
+		VALUE rb_headers = rb_hash_aref(rb_options, CSTR2SYM("headers"));
+		rugged_rb_ary_to_strarray(rb_headers, custom_headers);
+	}
+}
+
 static int parse_prune_type(VALUE rb_prune_type)
 {
 	if (rb_prune_type == Qtrue) {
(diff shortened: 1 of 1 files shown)


arthurschreiber:
@spraints Cool stuff!

As-is, this is going to leak memory on every call. You need call `git_strarray_free()` after you're done with the `git_strarray` struct.

You need to do something like this:

```c
static VALUE rb_git_remote_extra_http_headers(VALUE self)
{
	git_remote *remote;
	git_strarray http_headers;
	VALUE rb_result;


	Data_Get_Struct(self, git_remote, remote);

	rugged_exception_check(
		git_remote_extra_http_headers(remote, &http_headers)
	);

	rb_result = rugged_strarray_to_rb_ary(&http_headers);
	git_strarray_free(&http_headers);

	return rb_result;
}
```

spraints:
> As-is, this is going to leak memory on every call. You need call git_strarray_free() after you're done with the git_strarray struct.

Ah, cool. Good catch. libgit2/libgit2#3411 needs some more iteration. When that's more settled, I'll come back to this.

ext/rugged/rugged_remote.c
```diff
@@ -561,6 +561,17 @@ static VALUE rb_git_remote_fetch(int argc, VALUE *argv, VALUE self)
 	return rb_result;
 }
 
+void init_push_headers(VALUE rb_options, git_strarray *custom_headers)
+{
+	VALUE rb_headers;
```
arthurschreiber: You can move this declaration inside the `if` block.

ext/rugged/rugged_remote.c
```diff
@@ -561,6 +561,17 @@ static VALUE rb_git_remote_fetch(int argc, VALUE *argv, VALUE self)
 	return rb_result;
 }
 
+void init_push_headers(VALUE rb_options, git_strarray *custom_headers)
```
arthurschreiber: This can be `static`, I think.

ext/rugged/rugged_remote.c
spraints: > You can move this declaration inside the if block.

1324d66

ext/rugged/rugged_remote.c
spraints: TIL `static`. 4c741f6

arthurschreiber:
Looking great. :+1:

arthurschreiber:
@spraints Can you bump libgit2 and squash the other commits so this can be merged?

spraints:
> Can you bump libgit2 and squash the other commits so this can be merged?

Yep! fde2448 is squashed with an updated libgit2. :grinning:

Looking at the diff now, it looks like I could add support for custom fetch headers too. IIRC, I added that to https://github.com/libgit2/libgit2/pull/3411, but it isn't here. Would you like me to do that?

spraints:
> Looking at the diff now, it looks like I could add support for custom fetch headers too. IIRC, I added that to libgit2/libgit2#3411, but it isn't here. Would you like me to do that?

2f1f02d adds custom headers in more places. If it looks good, I can squash again.

arthurschreiber:
Oh, good catch with freeing of the strarray objects. Everything LGTM!

spraints:
> Oh, good catch with freeing of the strarray objects. Everything LGTM!

:metal: It's now squashed down to one commit. :smile_cat:

arthurschreiber:
PEEEEEEEEEEEEEEERFEEEEECT! :heart:

PR #521, opened 2015-09-16
Treebuilder remove return

albertsun:
The docs here say that with a Rugged::Tree::Builder object the remove method should return false if the file does not exist, but right now it'll actually raise a Rugged::TreeError.

Here's a test case and small fix for that.


arthurschreiber:
Cool, thanks! :sparkles:

Issue #524, opened 2015-09-29
Backends freed twice when failing to init


tpickett66:
While working on implementing a non-fs storage backend I ran into a situation where initing a repo will cause the storage backends to be freed twice when the backends initialize successfully but the first lookup agains the RefDB fails. It looks like the problem is in the cleanup routine [in  rugged_repo_new_with_backend in rugged_repo.c](https://github.com/libgit2/rugged/blob/master/ext/rugged/rugged_repo.c#L251-L252), libgit's free functions will free backends if they're attached to the db leading rugged to not need to call free. I'll put together a patch to fix the issue but I want to make sure I'm not totally misunderstanding something.

Code to reproduce the issue:
```ruby
# extconf.rb
require 'mkmf'

$CFLAGS << " #{ENV["CFLAGS"]}"
$CFLAGS << " -g"
$CFLAGS << " -O3" unless $CFLAGS[/-O\d/]
$CFLAGS << " -Wall -Wno-comment"

if !find_executable('cmake')
  abort "ERROR: CMake is required to build Rugged."
end

if !(MAKE = find_executable('gmake') || find_executable('make'))
  abort "ERROR: GNU make is required to build Rugged."
end

CWD = File.expand_path(File.dirname(__FILE__))
RUGGED_EXT_DIR = File.join(CWD, 'ext', 'rugged')
LIBGIT2_DIR = File.join(CWD, 'vendor', 'libgit2')

# Include rugged's header for the backend interface definition
puts "Using rugged headers from #{RUGGED_EXT_DIR}\n"
$CFLAGS << " -I#{RUGGED_EXT_DIR}"

# Link against rugged's libgit2
puts "Using libgit2 from #{LIBGIT2_DIR}/include and #{LIBGIT2_DIR}/build (rugged bundled)\n"
$CFLAGS << " -I#{LIBGIT2_DIR}/include"
$LIBPATH.unshift "#{LIBGIT2_DIR}/build"

create_makefile("partial_backend")
```

```c
// partial_backend.c
#include <stdio.h>
#include <ruby.h>
#include <git2.h>
#include <git2/sys/odb_backend.h>
#include <git2/sys/refdb_backend.h>
#include <rugged.h>

extern VALUE rb_mRugged;
extern VALUE rb_cRuggedBackend;

VALUE rb_cRuggedPartialBackend;

typedef struct {
  rugged_backend backend;
} rugged_bridge_backend;

typedef struct {
  git_odb_backend parent;
[...]

PR #525, opened 2015-09-29
Ensure DB backends haven't already been freed

tpickett66:
When DB backends are successfully attached to the DB instacnes they
get freed along with the DB so we don't have to. Also, the free method
on the backends is optional so we shouldn't assume it is present.

Fixes: #524

### DIFF
--- a/ext/rugged/rugged_repo.c
+++ b/ext/rugged/rugged_repo.c
@@ -216,8 +216,8 @@ static void rugged_repo_new_with_backend(git_repository **repo, VALUE rb_path, V
 
 	error = git_odb_add_backend(odb, odb_backend, 1);
 	if (error) {
-		if (odb_backend->free) odb_backend->free(odb_backend);
-		else git__free(odb_backend);
+		assert(odb_backend->free);
+		odb_backend->free(odb_backend);
 		goto cleanup;
 	}
 
@@ -229,8 +229,8 @@ static void rugged_repo_new_with_backend(git_repository **repo, VALUE rb_path, V
 
 	error = backend->refdb_backend(&refdb_backend, backend, path);
 	if (error) {
-		if (refdb_backend->free) odb_backend->free(odb_backend);
-		else git__free(refdb_backend);
+		assert(refdb_backend->free);
+		refdb_backend->free(refdb_backend);
 		goto cleanup;
 	}
 
--- a/ext/rugged/rugged_repo.c
+++ b/ext/rugged/rugged_repo.c
@@ -189,6 +189,8 @@ static void load_alternates(git_repository *repo, VALUE rb_alternates)
 static void rugged_repo_new_with_backend(git_repository **repo, VALUE rb_path, VALUE rb_backend)
 {
 	char *path;
+	int odb_associated = 0;
+	int refdb_associated = 0;
 
 	git_odb *odb = NULL;
 	git_odb_backend *odb_backend = NULL;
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Hey @tpickett66!

Thanks for your contribution! :smile: Before I can merge your patch, can you please change the indentation from spaces to tabs? Inside *.c files, we indent using tabs, the only exception being documentation blocks for rdoc.

I'll also have to take a quick look into the libgit2 source to verify this fix is correct, but other than these two things, this looks solid! :+1:

tpickett66:
Doh, I thought I set vim to tabs, I'll go and fix that :-)

arthurschreiber:
Actually, I think you can fix this without introducing new variables like this:

```diff
diff --git a/ext/rugged/rugged_repo.c b/ext/rugged/rugged_repo.c
index 15a086c..a16d098 100644
--- a/ext/rugged/rugged_repo.c
+++ b/ext/rugged/rugged_repo.c
@@ -215,7 +215,10 @@ static void rugged_repo_new_with_backend(git_repository **repo, VALUE rb_path, V
        if (error) goto cleanup;
 
        error = git_odb_add_backend(odb, odb_backend, 1);
-       if (error) goto cleanup;
+       if (error) {
+               odb_backend->free(odb_backend);
+               goto cleanup;
+       }
 
        error = git_repository_wrap_odb(repo, odb);
        if (error) goto cleanup;
@@ -224,7 +227,10 @@ static void rugged_repo_new_with_backend(git_repository **repo, VALUE rb_path, V
        if (error) goto cleanup;
 
        error = backend->refdb_backend(&refdb_backend, backend, path);
-       if (error) goto cleanup;
+       if (error) {
+               refdb_backend->free(refdb_backend);
+               goto cleanup;
+       }
 
        error = git_refdb_set_backend(refdb, refdb_backend);
        if (error) goto cleanup;
@@ -248,9 +254,6 @@ cleanup:
        git_odb_free(odb);
        git_refdb_free(refdb);
 
-       if (odb_backend != NULL) odb_backend->free(odb_backend);
-       if (refdb_backend != NULL) refdb_backend->free(refdb_backend);
-
        rugged_exception_check(error);
 }
```

What do you think?

tpickett66:
I've thought through it and can't see any problem with this approach beyond
`free ()` being optional, so we'll still want to check for that. I'll push
up anther commit in a bit.

arthurschreiber:
Mh, maybe I'm missing something, but I'm sure that the `free` methods on the backend are not really intended to be optional?

tpickett66:
I was confused by it too but it's [in the docs](https://libgit2.github.com/libgit2/#HEAD/type/git_refdb_backend) and `[odb_free](https://github.com/libgit2/libgit2/blob/master/src/odb.c#L595-L614)` checks that the function is set before deciding what to do. There's something similar for the [refdb](https://github.com/libgit2/libgit2/blob/master/src/refdb.c#L62-L70) too :-)

arthurschreiber:
Oh, that's... weird.

I think `free` being optional is an error on libgit2 side. I'll check this with the libgit2 people and get back to you asap.

Also, we can't use `git__free` here, because we don't know whether the objects we're freeing here were allocated with the same memory allocator that is used by libgit2 or not. (This is basically the same problem that is created by makeing `free` optional - libgit2 can't know what allocator was used to create the objects it's trying to free).

Thanks for your patience!

arthurschreiber:
See https://github.com/libgit2/libgit2/pull/3447. Can you remove the checks for the `free` struct member and remove the calls to `git__free`?

I can't imagine any sane refdb or odb backend that does not provide it's custom free function, and the code and documentation in libgit2 seem to be wrong.

Thanks again for the work here! :heart:

tpickett66:
No problem,  I'll make the adjustments after the kids go to sleep.

tpickett66:
I went ahead and added an `assert()` against the db backends supplying a `free()` implementation so the issue gets caught early.

arthurschreiber:
Thanks for the contribution! :heart:

PR #526, opened 2015-10-01
Adds Rugged::Remote#fetch(prune:)

pallan:
I am using Rugged quite a bit on a project and needed a method to prune the remotes. I didn't see a way to do it in the documentation so I thought I would give it a shot myself. Here is the result.  My C-fu has long since atrophied and I have never build a C-extension before so feel free to be harsh in your review/feedback.

It also looks like it is possible to pass a `prune` option to `git_remote_fetch` but I couldn't figure out how to do it yet.

### DIFF
--- a/ext/rugged/rugged_remote.c
+++ b/ext/rugged/rugged_remote.c
@@ -181,6 +181,19 @@ void rugged_remote_init_callbacks_and_payload_from_options(
 	}
 }
 
+static int parse_prune_type(VALUE rb_prune_type)
+{
+	if (rb_prune_type == Qtrue) {
+		return GIT_FETCH_PRUNE;
+	} else if (rb_prune_type == Qfalse) {
+		return GIT_FETCH_NO_PRUNE;
+	} else if (rb_prune_type == Qnil) {
+		return GIT_FETCH_PRUNE_UNSPECIFIED;
+	} else {
+		rb_raise(rb_eTypeError, "wrong argument type for :prune (expected true, false or nil)");
+	}
+}
+
 static void rb_git_remote__free(git_remote *remote)
 {
 	git_remote_free(remote);
@@ -502,6 +515,10 @@ static VALUE rb_git_remote_check_connection(int argc, VALUE *argv, VALUE self)
  *  :message ::
  *    The message to insert into the reflogs. Defaults to "fetch".
  *
+ *  :prune ::
+ *    Specifies the prune mode for the fetch. +true+ remove any remote-tracking references that
+ *    no longer exist, +false+ do not prune, +nil+ use configured settings Defaults to "nil".
+ *
  *  Example:
  *
  *    remote = Rugged::Remote.lookup(@repo, 'origin')
@@ -536,6 +553,10 @@ static VALUE rb_git_remote_fetch(int argc, VALUE *argv, VALUE self)
 		VALUE rb_val = rb_hash_aref(rb_options, CSTR2SYM("message"));
 		if (!NIL_P(rb_val))
 			log_message = StringValueCStr(rb_val);
+
+		VALUE rb_prune_type = rb_hash_aref(rb_options, CSTR2SYM("prune"));
+		if (!NIL_P(rb_prune_type))
+			opts.prune = parse_prune_type(rb_prune_type);
 	}
 
 	error = git_remote_fetch(remote, &refspecs, &opts, log_message);
@@ -620,6 +641,49 @@ static VALUE rb_git_remote_push(int argc, VALUE *argv, VALUE self)
 	return payload.result;
 }
 
+/*
+ *  call-seq:
+ *    remote.prune(options = {}) -> nil
+ *
+ *  Prune tracking refs that are no longer present on remote
+ *
+ *  Returns nil
+ *
+ *  The following options can be passed in the +options+ Hash:
+ *
+ *  :credentials ::
+ *    The credentials to use for the fetch operation. Can be either an instance of one
+ *    of the Rugged::Credentials types, or a proc returning one of the former.
+ *    The proc will be called with the +url+, the +username+ from the url (if applicable) and
+ *    a list of applicable credential types.
+ *
+ *  Example:
+ *
+ *    remote = Rugged::Remote.lookup(@repo, 'origin')
+ *    remote.prune()
+ *
+ */
+static VALUE rb_git_remote_prune(int argc, VALUE *argv, VALUE self)
+{
+	VALUE rb_options;
+	git_remote *remote;
+	git_fetch_options opts = GIT_FETCH_OPTIONS_INIT;
+	struct rugged_remote_cb_payload payload = { Qnil, Qnil, Qnil, Qnil, Qnil, Qnil, 0 };
+	int error;
+
+	Data_Get_Struct(self, git_remote, remote);
+	rb_scan_args(argc, argv, ":", &rb_options);
+	rugged_remote_init_callbacks_and_payload_from_options(rb_options, &opts.callbacks, &payload);
+
+	if ((error = git_remote_connect(remote, GIT_DIRECTION_FETCH, &opts.callbacks)) ||
+			(error = git_remote_prune(remote, &opts.callbacks)))
+	git_remote_disconnect(remote);
+
+	rugged_exception_check(error);
+
+	return Qnil;
+}
+
 void Init_rugged_remote(void)
 {
 	rb_cRuggedRemote = rb_define_class_under(rb_mRugged, "Remote", rb_cObject);
@@ -634,4 +698,5 @@ void Init_rugged_remote(void)
 	rb_define_method(rb_cRuggedRemote, "check_connection", rb_git_remote_check_connection, -1);
 	rb_define_method(rb_cRuggedRemote, "fetch", rb_git_remote_fetch, -1);
 	rb_define_method(rb_cRuggedRemote, "push", rb_git_remote_push, -1);
+	rb_define_method(rb_cRuggedRemote, "prune", rb_git_remote_prune, -1);
 }
--- a/ext/rugged/rugged_remote.c
+++ b/ext/rugged/rugged_remote.c
@@ -555,8 +555,7 @@ static VALUE rb_git_remote_fetch(int argc, VALUE *argv, VALUE self)
 			log_message = StringValueCStr(rb_val);
 
 		VALUE rb_prune_type = rb_hash_aref(rb_options, CSTR2SYM("prune"));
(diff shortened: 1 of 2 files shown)


ext/rugged/rugged_remote.c
```diff
@@ -620,6 +620,48 @@ static VALUE rb_git_remote_push(int argc, VALUE *argv, VALUE self)
...
+	git_remote *remote;
+	git_fetch_options opts = GIT_FETCH_OPTIONS_INIT;
+	struct rugged_remote_cb_payload payload = { Qnil, Qnil, Qnil, Qnil, Qnil, rb_hash_new(), 0 };
+	int error;
+
+	Data_Get_Struct(self, git_remote, remote);
+	rb_scan_args(argc, argv, "01", &rb_options);
+	rugged_remote_init_callbacks_and_payload_from_options(rb_options, &opts.callbacks, &payload);
+
+  if ((error = git_remote_connect(remote, GIT_DIRECTION_FETCH, &opts.callbacks)) ||
```
arthurschreiber: We're using tabs instead of spaces for indentation in c code.

ext/rugged/rugged_remote.c
```diff
@@ -620,6 +620,48 @@ static VALUE rb_git_remote_push(int argc, VALUE *argv, VALUE self)
...
+	    (error = git_remote_prune(remote, &opts.callbacks)))
+
```
arthurschreiber: This newline here can be removed (and is a bit confusing).

ext/rugged/rugged_remote.c
```diff
@@ -620,6 +620,48 @@ static VALUE rb_git_remote_push(int argc, VALUE *argv, VALUE self)
...
+  rugged_exception_check(error);
```
arthurschreiber: You need to disconnect the remote again if it was connected.

arthurschreiber:
Hey @pallan 

Thanks for this contribution! :heart: 

I pointed out some issues with the code as-is. Nothing too major, but these need to be resolved before I can go ahead and merge this.

pallan:
Thanks for the feedback @arthurschreiber. I have updated the code as recommended.

ext/rugged/rugged_remote.c
```diff
@@ -620,6 +620,49 @@ static VALUE rb_git_remote_push(int argc, VALUE *argv, VALUE self)
...
+
+	Data_Get_Struct(self, git_remote, remote);
+	rb_scan_args(argc, argv, "01", &rb_options);
+	rugged_remote_init_callbacks_and_payload_from_options(rb_options, &opts.callbacks, &payload);
+
+	if ((error = git_remote_connect(remote, GIT_DIRECTION_FETCH, &opts.callbacks)) ||
+			(error = git_remote_prune(remote, &opts.callbacks)))
+		rugged_exception_check(error);
+
+	git_remote_disconnect(remote);
```
arthurschreiber: You need to disconnect after the calls to `git_remote_connect` and `git_remote_prune`, but before the call to `rugged_exception_check`.

ext/rugged/rugged_remote.c
```diff
@@ -620,6 +620,49 @@ static VALUE rb_git_remote_push(int argc, VALUE *argv, VALUE self)
...
+static VALUE rb_git_remote_prune(int argc, VALUE *argv, VALUE self)
+{
+	VALUE rb_options;
+	git_remote *remote;
+	git_fetch_options opts = GIT_FETCH_OPTIONS_INIT;
+	struct rugged_remote_cb_payload payload = { Qnil, Qnil, Qnil, Qnil, Qnil, rb_hash_new(), 0 };
+	int error;
+
+	Data_Get_Struct(self, git_remote, remote);
+	rb_scan_args(argc, argv, "01", &rb_options);
```
arthurschreiber: This should be `rb_scan_args(argc, argv, ":", &rb_options);` instead.

ext/rugged/rugged_remote.c
```diff
@@ -620,6 +620,49 @@ static VALUE rb_git_remote_push(int argc, VALUE *argv, VALUE self)
...
+ *    remote = Rugged::Remote.lookup(@repo, 'origin')
+ *    remote.prune()
+ *
+ */
+static VALUE rb_git_remote_prune(int argc, VALUE *argv, VALUE self)
+{
+	VALUE rb_options;
+	git_remote *remote;
+	git_fetch_options opts = GIT_FETCH_OPTIONS_INIT;
+	struct rugged_remote_cb_payload payload = { Qnil, Qnil, Qnil, Qnil, Qnil, rb_hash_new(), 0 };
```
arthurschreiber: This should be `struct rugged_remote_cb_payload payload = { Qnil, Qnil, Qnil, Qnil, Qnil, Qnil, 0 };`. I don't think the hash object created here is used anywhere.

ext/rugged/rugged_remote.c
carlosmn: The disconnect should come as soon as possible. The last received references are kept in memory, so you don't need the socket to be open to read that, and it's good form not to keep system resources open longer than necessary (this is also the order in which `git_remote_fetch()` performs these operations).

arthurschreiber: @carlosmn You're totally right. I was more concerned with the fact that with the current code, if an error happens on `git_remote_prune`, the remote will never be disconnected.

arthurschreiber: Wait. Why is this connecting the remote at all? `git_remote_prune` is an offline operation, if I'm not totally wrong. No need to connect / disconnect the remote, at all.

carlosmn: It depends on whether you're thinking of the operation or the command. The operation just needs the remote to have connected (preferably recently) but `git remote prune` would connect. It depends which way you want to go. I thought you had noticed this difference, so I didn't mention it, but my preference would be to do it by acting on an existing remote.

However, `git_remote_prune()` is meant as a low-level operation, which `git_remote_fetch()` (or someone implementing their own version) can call. The recommended way to ask for a prune is to pass the option to `git_remote_fetch()` which would be `#fetch()` here. Connecting on `#prune()` when `git_remote_prune()` doesn't would be surprising as someone who uses multiple languages to talk to libgit2.

pallan: My first implementation did not connect to the remote and that resulted in an error that the remote has not connected yet.  I would prefer to implement this as `#fetch(prune: true)` or some such but haven't figured out how to make that work yet.

ext/rugged/rugged_remote.c
pallan: good point.

pallan:
I have updated the method based on supplied feedback.

ext/rugged/rugged_remote.c
carlosmn: Indeed, that's because it's an optional step during a fetch. The option to `#fetch` would be a tristate representing the field in the fetch optoins https://github.com/libgit2/libgit2/blob/master/include/git2/remote.h#L475-L488

pallan:
I added a `:prune` option to the `fetch` method.  The possible options and the pruning type they map to are, `true => GIT_FETCH_PRUNE, false => GIT_FETCH_NO_PRUNE, nil => GIT_FETCH_PRUNE_UNSPECIFIED`.  The default is `nil` which was the existing behaviour.

ext/rugged/rugged_remote.c
```diff
@@ -181,6 +181,17 @@ void rugged_remote_init_callbacks_and_payload_from_options(
...
 }
 
+static int parse_prune_type(VALUE rb_prune_type)
+{
+	if (rb_prune_type == Qtrue) {
+		return GIT_FETCH_PRUNE;
+	} else if (rb_prune_type == Qfalse) {
+		return GIT_FETCH_NO_PRUNE;
+	} else {
+		return GIT_FETCH_PRUNE_UNSPECIFIED;
```
vmg: This `else` branch isn't quite right. It will return `UNSPECIFIED` for literally any variable that isn't `true` or `false`. You need better type checking here. :)

pallan: `GIT_FETCH_PRUNE_UNSPECIFIED` is the default behaviour and it was intentional that I did it this way. If its preferred to type check and raise and error if the value is not `true / false / nil` then I will update it to work like that.

vmg: Yes, Rugged tries to be strict with type checking. A value of `"foobar"` should not result into a valid default because it's bogus.

pallan: :+1: sounds good, I will update the code.

pallan:
I have updated the code to check for `Qnil` and added a TypeError raise in event it does not meet the type requirements.

ext/rugged/rugged_remote.c
```diff
@@ -536,6 +553,10 @@ static VALUE rb_git_remote_fetch(int argc, VALUE *argv, VALUE self)
 		VALUE rb_val = rb_hash_aref(rb_options, CSTR2SYM("message"));
 		if (!NIL_P(rb_val))
 			log_message = StringValueCStr(rb_val);
+
+		VALUE rb_prune_type = rb_hash_aref(rb_options, CSTR2SYM("prune"));
+		if (!NIL_P(rb_prune_type))
```
pallan: with the `Qnil` check in the `parse_prune_type` i could remove this check. I went with the "don't call it if I don't have to approach". Any opinions on that?

ext/rugged/rugged_remote.c
```diff
@@ -620,6 +641,49 @@ static VALUE rb_git_remote_push(int argc, VALUE *argv, VALUE self)
...
+ *    The proc will be called with the +url+, the +username+ from the url (if applicable) and
+ *    a list of applicable credential types.
+ *
+ *  Example:
+ *
+ *    remote = Rugged::Remote.lookup(@repo, 'origin')
+ *    remote.prune()
+ *
+ */
+static VALUE rb_git_remote_prune(int argc, VALUE *argv, VALUE self)
```
arthurschreiber: Can we remove this method? IMHO it does not make too much sense in rugged, especially not with the `:prune` option you added.

ext/rugged/rugged_remote.c
arthurschreiber: Either remove the `NIL_P` check here, or more preferably, remove the `rb_prune_type == Qnil` from inside `parse_prune_type`.

ext/rugged/rugged_remote.c
pallan: Sure, if you don't think it adds any value I will remove it.

pallan:
Removed `Rugged::Remote#prune`

Issue #529, opened 2015-10-04
Q: How to walk the history of a single file with rugged?


matthiasbeyer:
Q: How to walk the history of a single file with rugged?

I cannot find an example for this in the README and I cannot a way how to do this in the docs, can you please provide an example?

arthurschreiber:
Unfortunately, libgit2 does not support filtering the history by filename natively. But you can check https://github.com/libgit2/rugged/issues/343, where an alternative is described.

matthiasbeyer:
Thank you. Time for a feature request! :smile: If you want, you can leave this open as a feature request!

Issue #343, opened 2014-03-26
Rugged::Walker vs git log performance


LightGuard:
I realize Rugged probably won't be as fast as git log, but currently the performance for the log on a particular file is horrible. Compare on your machine, for me (in a ruby benchmark 1000 times) I'm seeing differences between code to be double the time in rugged as shelling out to git log:

``` ruby
require 'rugged'
require 'yaml'
require 'open3'
require 'benchmark'

class GitCommands
  def commit_info(repo_root, file_path)
    o, _ = Open3.capture2(%Q[git --git-dir=#{repo_root}/.git log --date=short --format='- { !ruby/symbol author: %an, !ruby/symbol date: %ad, !ruby/symbol hash: %h, !ruby/symbol subject: %f }' -- #{file_path.to_s.sub(/#{repo_root}\//, '')}])
    YAML.parse o
  end

  def commit_info_rugged(repo_root, file_path)
    repo = Rugged::Repository.new(repo_root)
    walker = Rugged::Walker.new(repo)
    walker.sorting(Rugged::SORT_DATE)
    walker.push(repo.last_commit)
    walker.inject([]) do |a, c|
      if (c.diff(paths: [file_path.to_s.sub(/#{repo_root}\//, '')]).size > 0)
        a << {author: c.author, date: c.time, hash: c.oid, subject: c.message.split("\n").first}
      end
      a
    end
  end
end

git = GitCommands.new
repo = '';
file = ''
amount = 1000

Benchmark.bmbm(22) do |x|
  x.report('rugged commit_info') { amount.times do ; git.commit_info_rugged(repo, file) end }

  x.report('git commit_info') { amount.times do ; git.commit_info(repo, file) end }
end
```

obviously replace repo and file :)

Numbers from my machine:

```
Rehearsal ----------------------------------------------------------
rugged commit_info      23.840000   2.550000  26.390000 ( 26.793017)
git commit_info          1.680000   1.290000  15.260000 ( 15.066757)
------------------------------------------------ total: 41.650000sec

                             user     system      total        real
rugged commit_info      23.890000   2.580000  26.470000 ( 26.858275)
git commit_info          1.700000   1.340000  16.260000 ( 15.967883)
```

connorshea:
Was this ever fixed?

carlosmn:
There isn't really anything _to_ fix. The revwalk and git-log with path simplification are different beasts. When you simplify by path, you use the diff information to tell you down which sub-histories you don't need to go down.

[...]

PR #531, opened 2015-10-05
[WIP] Add method to get history of a specific file

matthiasbeyer:
This is a naive implementation of a "get history of a single file" (as discussed in #529 and shown in #343) and I'm seeking for feedback here.

I didn't add tests because I don't know how to approach this and I'd be happy about improvement suggestions and the like!

### DIFF
--- a/lib/rugged/repository.rb
+++ b/lib/rugged/repository.rb
@@ -223,5 +223,28 @@ module Rugged
 
       remote_or_url.push(*args)
     end
+
+    # Get the history of a file
+    #
+    # path - Path of the file
+    #
+    # Returns an Array of Hashes, each Hash describing a commit with author,
+    # date, hash and subject.
+    def history_of_file path
+      w = Rugged::Walker.new self
+      w.sorting Rugged::SORT_DATE
+      w.push self.last_commit
+      w.inject [] do |a, c|
+        if c.diff(paths: [path.to_s.sub(/#{repo_root}\//, '')]).size > 0
+          a << {
+            author:   c.author,
+            date:     c.time,
+            hash:     c.oid,
+            subject:  c.message.split("\n").first
+          }
+        end
+        a
+      end
+    end
   end
 end


arthurschreiber:
Hey @matthiasbeyer

Thanks for the contribution! I'm not sure how to proceed here, as this is something we want in libgit2, and not come up with our own implementation in Rugged. Also, your solution, as you pointed out, is quite naive and does not provide more advanced functionality that `git log` would support (like rename detection). I'd rather not take in an implementation that'll later differ from what libgit2 is going to provide in the future.

matthiasbeyer:
Hi!

Are there plans for this in libgit2? So well... Then I'll implement it in my application until it is provided through libgit2, of course!

Anyways, this is a really nice library and I really enjoy using it. The docs could use some polishing, but well... nothing is perfect! Thanks for developing it!

arthurschreiber:
@carlosmn has been talking about implementing this for a long time now, but the problem is that a "correct" implementation will be everything but trivial.

There's a feature in git that is eliminiating intermediate commits if you do a git log filtered by a path, and libgit2 will have to implement something similar to match the results of git. Unfortunately, libgit2 can't take the same approach that git is using to perform this elimination.

But I can promise you once it has landed in libgit2, it will also be available through rugged. :smile:

carlosmn:
I've largely been thinking about how inappropriate the current API is to return this kind of data. There is libgit2/libgit2#3041 to track which features we want to add. The revwalk itself just isn't enough to do anything with history simplification, which path filtering would need.

Also note that while giving a list of commits which touch a file might be what you want to do sometimes, this is not enough to provide an API we'd be happy with. With an implementation such as this, you do not have the information necessary for something like `git log --graph`. You may want to throw it away, but rugged should be providing it.

There's also the issue that "history_of_file" is an inaccurate way to describe what's happening. The file doesn't exist as an entity. What this is doing is filtering by the path. If a file at a particular path is removed and then re-appears, we don't know that it's the same file (and e.g. Subversion would consider this to be two different files, whereas git would show it).

PR #534, opened 2015-10-06
XDiff upstream changes

vmg:
Tip of `master` for libgit2 with the security fixes for XDiff.


arthurschreiber:
Looking good. :+1:

PR #542, opened 2015-10-28
Experimental fast merges branch

vmg:
Some crazy ass shit here with the new pooled allocator and the `skip_reuc` flag. Let's try it ouuttttt


arthurschreiber:
:+1:

PR #549, opened 2015-12-01
Add support for new GIT_MERGE_NO_RECURSIVE option for merges 

brianmario:
This bumps libgit2 to a589f22a9441523327e4c987c5b5f5f857472067, fixes Rugged for the API changes from that bump and adds support for the `GIT_MERGE_NO_RECURSIVE` tree merge option.

/cc @josh @aroben @vmg @ethomson

### DIFF
--- a/ext/rugged/rugged_tree.c
+++ b/ext/rugged/rugged_tree.c
@@ -614,15 +614,15 @@ void rugged_parse_merge_options(git_merge_options *opts, VALUE rb_options)
 		}
 
 		if (RTEST(rb_hash_aref(rb_options, CSTR2SYM("renames")))) {
-			opts->tree_flags |= GIT_MERGE_TREE_FIND_RENAMES;
+			opts->flags |= GIT_MERGE_FIND_RENAMES;
 		}
 
 		if (RTEST(rb_hash_aref(rb_options, CSTR2SYM("fail_on_conflict")))) {
-			opts->tree_flags |= GIT_MERGE_TREE_FAIL_ON_CONFLICT;
+			opts->flags |= GIT_MERGE_FAIL_ON_CONFLICT;
 		}
 
 		if (RTEST(rb_hash_aref(rb_options, CSTR2SYM("skip_reuc")))) {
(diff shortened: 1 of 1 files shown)


ext/rugged/rugged_tree.c
```diff
@@ -614,15 +614,19 @@ void rugged_parse_merge_options(git_merge_options *opts, VALUE rb_options)
...
+			opts->flags |= GIT_MERGE_FAIL_ON_CONFLICT;
 		}
 
 		if (RTEST(rb_hash_aref(rb_options, CSTR2SYM("skip_reuc")))) {
-			opts->tree_flags |= GIT_MERGE_TREE_SKIP_REUC;
+			opts->flags |= GIT_MERGE_SKIP_REUC;
+		}
+
+		if (RTEST(rb_hash_aref(rb_options, CSTR2SYM("no_recursive")))) {
+			opts->flags |= GIT_MERGE_NO_RECURSIVE;
```
brianmario: Is recursive the _new_ default? If so, maybe I should name the new Ruby option `recursive`, where `GIT_MERGE_NO_RECURSIVE` would be set by default and preserve the existing behavior?

carlosmn: It's the git.git default, so it's the default here to match the expectation. It's a *new* default because the feature was missing before now.

vmg: `no_recursive` makes sense as a flag. We'll leave it like this and add the flag by default to the options in GitRPC.

vmg:
This is real fucking neat. I was going to do this myself just today. Thanks for taking care of it. :)

ext/rugged/rugged_tree.c
arthurschreiber: > no_recursive makes sense as a flag

Personally, I'd have preferred something like `recursive: true` being the default (so you can say `recursive: false`). `no_recursive: true` feels a bit convoluted. On the other hand, following the libgit2 names is not a bad idea either. Bikeshedding galore. :worried:

Issue #506, opened 2015-08-06
undefined reference to WinHttpReadData


jweather:
Building master on Windows Server 2012 64-bit with MSVC 2013, Ruby 2.1.6, DevKit, and CMake 3.3.0:

`C:/Users/Administrator/Desktop/spi2/app/server/native/windows/bin/ruby.exe extconf.rb
checking for gmake... no
checking for make... yes
checking for cmake... yes
checking for pkg-config... yes
 -- cmake .. -DBUILD_CLAR=OFF -DTHREADSAFE=ON -DBUILD_SHARED_LIBS=OFF -DCMAKE_C_FLAGS=-fPIC -DCMAKE_BUILD_TYPE=RelWithDebInfo -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX=C:/
 -- C:\Users\Administrator\Desktop\devkit\bin/make.exe
checking for main() in -lgit2... yes
checking for git2.h... yes
creating Makefile

make "DESTDIR=" clean

make "DESTDIR="
generating rugged-i386-mingw32.def
compiling rugged.c
compiling rugged_backend.c
compiling rugged_blame.c
compiling rugged_blob.c
compiling rugged_branch.c
compiling rugged_branch_collection.c
compiling rugged_commit.c
compiling rugged_config.c
compiling rugged_cred.c
compiling rugged_diff.c
compiling rugged_diff_delta.c
compiling rugged_diff_hunk.c
compiling rugged_diff_line.c
compiling rugged_index.c
compiling rugged_note.c
compiling rugged_object.c
compiling rugged_patch.c
compiling rugged_reference.c
compiling rugged_reference_collection.c
compiling rugged_remote.c
compiling rugged_remote_collection.c
compiling rugged_repo.c
compiling rugged_revwalk.c
compiling rugged_settings.c
compiling rugged_signature.c
compiling rugged_submodule.c
compiling rugged_submodule_collection.c
compiling rugged_tag.c
compiling rugged_tag_collection.c
compiling rugged_tree.c
linking shared-object rugged/rugged.so
C:/Users/Administrator/Desktop/spi2/app/server/native/windows/lib/ruby/gems/2.1.0/gems/rugged-0.23.0/ext/rugged/../../vendor/libgit2/build/libgit2.a(winhttp.c.obj): In function `winhttp_stream_close':
[...]

PR #554, opened 2015-12-18
Build on Windows with WinHTTP

carlosmn:
Insert the right libraries to link to to get WinHTTP working. These are
normally done by the build system, but we're building statically so we put
them in ourselves.

We don't write out any dependencies to pkg-config, so let's get rid of it.
To boot, it's not always installed on this platform.

---

It'd be nice to be able to build this without having to specify the dependencies manually, like we do for the other deps on unix, but pkg-config isn't really a thing on Windows (and it doesn't come in the devkit), so we kinda have to put the dependencies here.

This resolves the linking issues from #506 though the tests still fail spectacularly as files are either missing or read-only. But those are all issues with the teardown, so the actual tests are working.

### DIFF
--- a/ext/rugged/extconf.rb
+++ b/ext/rugged/extconf.rb
@@ -15,6 +15,10 @@ def sys(cmd)
   ret
 end
 
+def on_windows
+  RbConfig::CONFIG['host_os'] =~ /mswin|mingw/
+end
+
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Thanks @carlosmn!

PR #557, opened 2016-01-03
Only ask pkg-config for -l options

mscharley:
On Windows the default configuration for CMake ends up causing `pkg-config` to generate the following options:

```
# libgit2.pc
libdir=C:/Program Files (x86)/libgit2/lib
includedir=C:/Program Files (x86)/libgit2/include

Name: libgit2
Description: The git library, take 2
Version: 0.23.3

Libs: -L${libdir} -lgit2
Libs.private: 
Requires.private: 

Cflags: -I${includedir}
```

```
C:\Users\Matthew $ pkg-config --libs --static C:\Ruby22-x64\lib\ruby\gems\2.2.0\gems\rugged-0.23.3\vendor\libgit2\build\libgit2.pc
-LC:/Program Files (x86)/libgit2/lib -lgit2
```

This causes quoting issues because of the spaces. Since rugged never actually uses the installation folder anyway since we're building something to vendor it makes sense to just not ask `pkg-config` for the -L options.

### DIFF
--- a/ext/rugged/extconf.rb
+++ b/ext/rugged/extconf.rb
@@ -78,7 +78,7 @@ else
         $LIBS << " -lwinhttp -lcrypt32 -lrpcrt4 -lole32"
       else
         pcfile = File.join(LIBGIT2_DIR, "build", "libgit2.pc")
-        $LDFLAGS << " " + `pkg-config --libs --static #{pcfile}`.strip
+        $LDFLAGS << " " + `pkg-config --libs-only-l --static #{pcfile}`.strip
       end
     end
   end


mscharley:
I just realised that recent beta-only changes have rendered this unneeded for Windows, though could still (unlikely) avoid some issues on *NIX as well.

arthurschreiber:
> I just realised that recent beta-only changes have rendered this unneeded for Windows

I was about to tell you exactly that. :smile:

I think we used the `--libs-only-l` option before, but that actually caused some libraries not to be found (I think that was on OS X), and thus this was changed to be only `--libs`. I hope it's okay if I close this PR without merging. If you actually run into an issue with the `--libs` option outside of windows, feel free to let us know! Thanks! :heart:

mscharley:
I feel a write-up coming on. Rugged was fairly straightforward this bug not withstanding, but libssh2 is being annoying. This stuff's never straightforward on Windows sadly. Cheers for looking anyway! :+1: Just glad it's fixed.

PR #562, opened 2016-01-15
Add Commit#header_field method

mastahyeti:
This adds `Commit#header_field`, which was added to libgit2 in https://github.com/libgit2/libgit2/pull/3240. This should make it possible to fetch and verify commit signatures in Ruby.

/cc @carlosmn @peff

### DIFF
--- a/ext/rugged/rugged_commit.c
+++ b/ext/rugged/rugged_commit.c
@@ -23,6 +23,7 @@
  */
 
 #include "rugged.h"
+#include "git2/commit.h"
 
 extern VALUE rb_mRugged;
 extern VALUE rb_cRuggedObject;
@@ -576,6 +577,20 @@ static VALUE rb_git_commit_header_field(VALUE self, VALUE rb_field) {
 	return rb_result;
 }
 
+static VALUE rb_git_commit_header(VALUE self) {
+	VALUE rb_result;
+	git_commit *commit;
+	const char *raw_header;
+
+	Data_Get_Struct(self, git_commit, commit);
+
+	raw_header = git_commit_raw_header(commit);
+
+	rb_result = rb_enc_str_new(raw_header, strlen(raw_header), rb_utf8_encoding());
+
+	return rb_result;
+}
+
 void Init_rugged_commit(void)
 {
 	rb_cRuggedCommit = rb_define_class_under(rb_mRugged, "Commit", rb_cRuggedObject);
@@ -600,4 +615,5 @@ void Init_rugged_commit(void)
 	rb_define_method(rb_cRuggedCommit, "to_mbox", rb_git_commit_to_mbox, -1);
 
 	rb_define_method(rb_cRuggedCommit, "header_field", rb_git_commit_header_field, 1);
+	rb_define_method(rb_cRuggedCommit, "header", rb_git_commit_header, 0);
 }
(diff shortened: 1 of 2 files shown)


mastahyeti:
I guess to verify a signature, you need to get the whole header section. So another method for that might be useful too.

peff:
@vmg is probably a better choice for reviewing rugged changes.

But yeah, if you want to actually verify a signature you need to get the whole thing, byte for byte minus the `gpgsig` headers. The relevant git code is [here](https://github.com/gitster/git/blob/fc10eb5b87a75af2cb93a3932897572f2c544915/commit.c#L1124-L1162).

mastahyeti:
I added a `Commit#header` method as well.

mastahyeti:
Looks like `@repo.read(oid).data` is the signing payload I'm looking for. Let me know if I should revert 8454965, or if `Commit#header` is useful for other things.

peff:
Wouldn't `@repo.read(oid).data` have all of the headers in it, including the detached signature? The signature is over everything _except_ the `gpgsig` headers. You can rip them out yourself, of course, but it might be nice if the library has some support.

mastahyeti:
> You can rip them out yourself, of course, 

Yep. That was my plan.

> but it might be nice if the library has some support.

I'd be happy to implement that in rugged instead if others would find it useful.

arthurschreiber:
@mastahyeti Both changes look good to me.

> I'd be happy to implement that in rugged instead if others would find it useful.

If you can find a good name for that method, I'd be happy to merge it into rugged. :smile: 

I can also push a new beta release with these changes if you want them to be available soon.

carlosmn:
Long-term the library should support doing a lot of this itself since the gpg signature is the standard way to perform this validation; and doing it by hand is really annoying (especially the signing part). We'd probably start with something like `git_commit_read_signed()` or similar to extract the gpg signature and present the buffer without that field so the user can verify themselves. Eventually we'd want to optionally link to `libgpg` so you can ask the library to verify or sign commits.

vmg:
I think @carlosmn is right. We don't want to duplicate the parsing logic anywhere. @mastahyeti can we hold this until Feb 2nd when Carlos starts full time? It will be an interesting project for him to work on.

vmg:
(This PR looks OK, by the way, so feel free to merge it although --again-- the actual signature extraction should be performed in C eventually)

mastahyeti:
> I think @carlosmn is right. We don't want to duplicate the parsing logic anywhere. @mastahyeti can we hold this until Feb 2nd when Carlos starts full time? It will be an interesting project for him to work on.

:+1: The `git_commit_read_signed` method should be pretty simple to implement in Ruby, so I might do that for now with the expectation that @carlosmn will reimplement it in C once he starts full time.

mastahyeti:
Would it be okay to merge this? I'd like to use these methods for now while we wait for a C implementation.

arthurschreiber:
@mastahyeti Do you need this to be tagged?

mastahyeti:
Yeah, that would be helpful. Thanks!

arthurschreiber:
@mastahyeti Tagged and pushed as `0.24.0b12`.

mastahyeti:
:bow:

PR #563, opened 2016-02-02
Make Commit#header_field return nil for non-existent headers

mastahyeti:
The `Commit#header_field` method added in #562 raises an exception currently if the header field is not present. This PR changes the method to return `nil` in this case.


brianmario:
Looks good to me!

PR #565, opened 2016-02-03
Add Commit#header_field? for checking header existence

mastahyeti:
This PR adds a `Commit#header_field?` method for checking if a commit has the given header.


brianmario:
This looks good as well

mastahyeti:
Thanks @brianmario. Can you cut a new tag so we can start using this and #563?

PR #567, opened 2016-02-08
commit: Cleanup header extraction APIs

vmg:
Just some simple changes since I missed review the previous time.
`Commit#header_field?` can be implemented directly in Ruby since the C
code is almost 100% redundant.

The original `Commit#header_field` has been refactored to prevent a
possible memory leak, and to properly tag the resulting header with the
encoding of the commit message, if we know it.

There are no breaking API changes

cc @mastahyeti

### DIFF
--- a/ext/rugged/rugged_commit.c
+++ b/ext/rugged/rugged_commit.c
@@ -563,57 +563,36 @@ static VALUE rb_git_commit_to_mbox(int argc, VALUE *argv, VALUE self)
  *
  *  Returns +commit+'s header field value.
  */
-static VALUE rb_git_commit_header_field(VALUE self, VALUE rb_field) {
+static VALUE rb_git_commit_header_field(VALUE self, VALUE rb_field)
+{
 	git_buf header_field = { 0 };
+	git_commit *commit = NULL;
+
+	const char *encoding_name;
+	rb_encoding *encoding = rb_utf8_encoding();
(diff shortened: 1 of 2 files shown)


vmg:
@arthurnn: Travis is barking in 2 builds because of "missing compiler tools". Do you have some cycles to look into it?

mastahyeti:
:metal:

arthurnn:
@vmg only because you're asking ❤️

vmg:
Huuuuuh that was actually for @arthurschreiber. Sorry @arthurnn! :)))

aroben:
Part of the idea of `header_field?` was to avoid a string allocation when just checking whether a header exists. E.g., perhaps you have a bunch of commits and want to see which of them have a signature header.

vmg:
@aroben: I see. Do you think it's worth it? The Ruby string allocation is noise compared to the cost of parsing and allocating the header in libgit2, which always happens regardless of the chosen API. That's why I refactored the new API.

aroben:
I leave that determination to you. I don't know much about this code.

carlosmn:
I'd go with performing this smallish extra allocation until we actually see this be a non-trivial part of the runtime. There's more things we can do like accepting a commit id so we avoid parsing this into a commit object, which is bound to be more expensive overall than this malloc.

PR #568, opened 2016-02-09
Add Commit::extract_signature

carlosmn:
This returns the signature and the signed data so the user can compare
them.

---

This depends on libgit2/libgit2#3599 so we shouldn't merge until that gets resolved.

### DIFF
--- a/ext/rugged/rugged_commit.c
+++ b/ext/rugged/rugged_commit.c
@@ -612,11 +612,58 @@ static VALUE rb_git_commit_header(VALUE self)
 	return rb_str_new_utf8(raw_header);
 }
 
+/*
+ *  call-seq:
+ *    Rugged::Commit.extract_signature(repo, commit, field_name) -> [str, str]
+ *
+ *  Returns +commit+'s signature in 'field' and the signed data
+ *
+ *  The signature is done over the contents of the commit without the
+ *  signature block in the header, which is the data in the second
+ *  element in the return array.
+ */
+static VALUE rb_git_commit_extract_signature(int argc, VALUE *argv, VALUE self)
+{
+	int error;
+	VALUE ret_arr;
+	git_oid commit_id;
+	const char *field;
+	git_repository *repo;
+	git_buf signature = {0}, signed_data = {0};
+	VALUE rb_repo, rb_commit, rb_field = Qnil;
+
+	rb_scan_args(argc, argv, "21", &rb_repo, &rb_commit, &rb_field);
+
+	rugged_check_repo(rb_repo);
+	Data_Get_Struct(rb_repo, git_repository, repo);
+
+	error = git_oid_fromstr(&commit_id, StringValueCStr(rb_commit));
+	rugged_exception_check(error);
+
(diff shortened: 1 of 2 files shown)


ext/rugged/rugged_commit.c
```diff
@@ -612,11 +612,53 @@ static VALUE rb_git_commit_header(VALUE self)
...
+
+	error = git_commit_extract_signature(&signature, &signed_data, repo, &commit_id, StringValueCStr(rb_field));
+	if (error < 0) {
+		git_buf_free(&signature);
+		git_buf_free(&signed_data);
+	}
+
+	rugged_exception_check(error);
+
+	ret_arr = rb_ary_new3(2, rb_enc_str_new(signature.ptr, signature.size, rb_utf8_encoding()),
```
arthurschreiber: Shouldn't this use `git_commit_message_encoding` instead of hardcoding UTF-8?

vmg: actually, since we're meant to verify these bytes, they'd be better off as a binary/ascii string instead.

carlosmn: Yeah, they're really just bytes. Is there a better type to use for this? The message encoding is only really supposed to be about the commit message, even if we end up using it for quite a few more things.

vmg: Yeah, tag it as `ASCII_8BIT`, that's the default "binary" encoding in Ruby.

carlosmn: Changed the encoding to `rb_ascii8bit_encoding()` which should be the right one. We don't seem to use it anywhere else in the bindings, which seems odd, though;

vmg: Hm. We should be using it. Let me double check.

carlosmn: For the blob we use `rb_str_new()` which seems to do the same thing with fewer chars, so maybe use that instead of specifying it?

vmg: Right. That's what I was missing. Let's go with `rb_str_new`.

brianmario:
@carlosmn I bumped libgit2 to pull in the new call you added, things appear to be working as expected. Was there anything else you wanted to get in here before merging? Possibly an instance method on `Rugged::Commit` for this that wraps this one but with a single [optional] `field` string or something?

arthurschreiber:
@carlosmn The commit (and PR title) refer to `extract_header`, while the method is called `extract_signature`. I guess the latter is correct?

carlosmn:
> The commit (and PR title) refer to extract_header, while the method is called extract_signature. I guess the latter is correct?

Yeah, I often think I could do with copy editor :)

>Possibly an instance method on Rugged::Commit for this that wraps this one but with a single [optional] field string or something?

We could do with an instance method, but it'd just be `def extract_signature; Commit::extract_signature(self.hex); end` so I'm a bit meh. I do still want to add  a `field` param to the static method (class method?) which was a change made after I wrote this version.

carlosmn:
Changed the commit message, added the optional signature argument and rebased so the order makes sense squashing these together. Should be ready for merging now.

PR #570, opened 2016-02-10
Add Tag.extract_signature method

mastahyeti:
This adds `Rugged::Tag.extract_signature method` to extract tag signatures/signing-data from tags.

/cc @vmg @carlosmn @brianmario

### DIFF
--- a/lib/rugged/tag.rb
+++ b/lib/rugged/tag.rb
@@ -1,5 +1,24 @@
 module Rugged
   class Tag < Rugged::Reference
+    GPG_SIGNATURE_PREFIX = "-----BEGIN PGP SIGNATURE-----".freeze
+
+    def self.extract_signature(repo, oid, prefix=GPG_SIGNATURE_PREFIX)
+      object = repo.read(oid)
+
+      unless object.type == :tag
+        raise GitRPC::InvalidObject, "Invalid object type #{object.type}, expected tag"
+      end
+
+      unless index = object.data.index(prefix)
+        raise GitRPC::InvalidObject, "Tag does not contain signature"
+      end
+
+      [
+        object.data.byteslice(index..-1),
+        object.data.byteslice(0...index)
+      ]
+    end
+
     def name
       canonical_name.sub(%r{^refs/tags/}, "")
     end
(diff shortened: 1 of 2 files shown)


lib/rugged/tag.rb
```diff
@@ -1,5 +1,24 @@
...
+
+    def self.extract_signature(repo, oid, prefix=GPG_SIGNATURE_PREFIX)
+      object = repo.read(oid)
+
+      unless object.type == :tag
+        raise GitRPC::InvalidObject, "Invalid object type #{object.type}, expected tag"
+      end
+
+      unless index = object.data.index(prefix)
+        raise GitRPC::InvalidObject, "Tag does not contain signature"
```
mastahyeti: Not sure what exceptions we should be raising here and above.

vmg:
Really feels like this should be at the libgit2 level. Any reason why you wrote it in Ruby?

mastahyeti:
I was talking with @carlosmn yesterday and he suggested in be done in Ruby in Rugged because it's such a simple string operation.

carlosmn:
Splitting this text into signature + signed data doesn't require any knowledge about Git int he code. You just take a buffer and split it where the signature starts. So the code is actually a lot clearer when we can use ruby rather than C to split it up.

If we notice it's too slow here, or more bindings end up implementing this, or libgit2 starts doing signature verification I'd put it in C, but meanwhile I think it makes more sense to use an expressive language.

vmg:
OK, makes sense to me.

lib/rugged/tag.rb
vmg: I think here you want to return `nil`, instead of raising. An exception is something exceptional. It would be more idiomatic to return nothing if there's nothing to return.

brianmario:
I'm a little indifferent about returning nil vs raising, but leaning more toward returning nil.

But this looks good to me other than that :+1:

mastahyeti:
I would lean towards `nil` as well, but wanted it to behave similarly to `Commit.extract_signature`, which raises.

vmg:
@mastahyeti: Let's make it return `nil` too. I'll get that one, you can change this PR. :ok_hand:

PR #572, opened 2016-02-11
Introduce rebasing

carlosmn:
Introduce the Rugged::Rebase and Rugged::RebaseOperation types to expose
the library's rebasing capabilities.

---

No tests yet, just wanna get the overall feature out there, since I think this is the first time I'm not just tweaking the existing types.

One thing I'm not quite sure about is what to do with `git_rebase_operation`. It's a very simple struct, so maybe we can put that into ruby code and initialize it from the C code. Does that work? Does MRI memoize the symbols I'm returning for the type or should we store them? Should we use subclassing here instead of a type?


ext/rugged/rugged_rebase.c
```diff
@@ -0,0 +1,206 @@
...
+	git_rebase_options options = GIT_REBASE_OPTIONS_INIT;
+
+	rb_scan_args(argc, argv, "21:", &rb_repo, &rb_branch, &rb_upstream, &rb_onto, &rb_options);
+	Data_Get_Struct(rb_repo, git_repository, repo);
+	Check_Type(rb_branch, T_STRING);
+	Check_Type(rb_upstream, T_STRING);
+
+	rugged_exception_check(git_oid_fromstr(&id, StringValueCStr(rb_branch)));
+	rugged_exception_check(git_annotated_commit_lookup(&branch, repo, &id));
+	rugged_exception_check(git_oid_fromstr(&id, StringValueCStr(rb_upstream)));
```
vmg: Memory leak. We need better handling for the error case or all the looked up commits will leak.

carlosmn:
We shouldn't be leaking anymore, and now all the functions are wrapped. Whenever we try to finish or abort the operation, we get a really weird error message, which looks like it may actually be an error string frm something else, but I'll have to investingate.

`#commit` is funny because the compulsory argument is the second one... I guess I'll have to move to one compulsory and two optional args.

arthurschreiber:
Woo, you're on :fire:!

arthurschreiber:
Regarding `git_rebase_operation`: Is there any reason why you can't convert this into a simple Hash instead of looking up these values for `id` and so on on-demand?

ext/rugged/rugged_rebase.c
```diff
@@ -0,0 +1,263 @@
...
+	Check_Type(rb_upstream, T_STRING);
+
+	if ((error = git_oid_fromstr(&id, StringValueCStr(rb_branch))) < 0 ||
+	    (error = git_annotated_commit_lookup(&branch, repo, &id)) < 0 ||
+	    (error = git_oid_fromstr(&id, StringValueCStr(rb_upstream))) < 0 ||
+	    (error = git_annotated_commit_lookup(&upstream, repo, &id)) < 0)
+		goto cleanup;
+
+	if (!NIL_P(rb_onto)) {
+		Check_Type(rb_onto, T_STRING);
```
arthurschreiber: You need to move all `Check_Type` calls (actually, everything that can raise a ruby exception) before any potential libgit2 calls that allocate memory, otherwise this code will leak if the `onto` parameter is passed but is something other than a string.

ext/rugged/rugged_rebase.c
```diff
@@ -0,0 +1,263 @@
...
+		if ((error = git_oid_fromstr(&id, StringValueCStr(rb_onto))) < 0 ||
+		    (error = git_annotated_commit_lookup(&onto, repo, &id)) < 0)
+			goto cleanup;
+	}
+
+	parse_rebase_options(&options, rb_options);
```
arthurschreiber: You need to move this in front of the other libgit2 calls as well, as `parse_rebase_options` can potentially raise a ruby exception.

ext/rugged/rugged_rebase.c
```diff
@@ -0,0 +1,263 @@
...
+		message = StringValueCStr(rb_message);
+	}
+
+	if (NIL_P(rb_committer))
+		rb_raise(rb_eArgError, "Expected non-nil committer");
+	else
+		committer = rugged_signature_get(rb_committer, NULL);
+
+	if (!NIL_P(rb_author))
+		author = rugged_signature_get(rb_author, NULL);
```
arthurschreiber: I know this is not related to your code and most likely exists in other areas of Rugged, but `comitter` will leak if `rugged_signature_get` raises a ruby exception.

ext/rugged/rugged_rebase.c
```diff
@@ -0,0 +1,263 @@
...
+	return Qnil;
+}
+
+static VALUE rb_git_rebase_finish(VALUE self, VALUE rb_sig)
+{
+	git_rebase *rebase;
+	git_signature *sig;
+
+	Data_Get_Struct(self, git_rebase, rebase);
+	sig = rugged_signature_get(rb_sig, NULL);
```
arthurschreiber: I think you need to free `sig` after using it.

ethomson:
I was overjoyed to wake up and see that you had bound this!  :dancers:  Note that I updated https://github.com/libgit2/libgit2/pull/3610 to include `git_merge_options` as well.

ext/rugged/rugged_rebase.c
carlosmn: Yeah, I'm pretty sure we leak *a lot* since those kinds of checks are everywhere.

arthurschreiber: Yeah. Unfortunately, exception handling in the MRI API is really a pain. :cry:

ext/rugged/rugged_rebase.c
carlosmn: Moved this and the other type check to the front.

ext/rugged/rugged_rebase.c
carlosmn: I wonder if it's realistic to have these functions return an exception so we can check it for non-null and perform cleanup before raising, like we jump to a `cleanup` label in libgit2 before returning the stored error code.

arthurschreiber: Yes, I've been thinking about this today after my first comment. If you want to catch a Ruby Exception in C-Land, you need to wrap the call using `VALUE rb_protect(VALUE (*)(VALUE), VALUE, int*);`.

We could add wrapper functions for things like `rugged_signature_get` that take all arguments, stash them into a struct, pass this down `rb_protect` with another function that unwraps that struct and calls the original `rugged_signature_get` again.

But that feels really convoluted and wasteful. You might apply some macro magic to make this look nicer, but the wrapping/unwrapping would still be there.

Another approach would be to split functions like `rugged_signature_get` into 2 separate functions like `rugged_signature_check` (which performs all the checks) and `rugged_signature_get` (which performs none of the checks and only handles conversion).

I think I favor the latter approach.

carlosmn: With the amount of `Check_Type()` around, the second approach does sound a lot more feasible. Otherwise we'd be wrapping absolutely everything ever.

ext/rugged/rugged_rebase.c
```diff
@@ -0,0 +1,273 @@
...
+
+	if (NIL_P(rb_options))
+		return;
+
+	val = rb_hash_aref(rb_options, CSTR2SYM("quiet"));
+	if (RTEST(val))
+		ret->quiet = 1;
+
+	val = rb_hash_aref(rb_options, CSTR2SYM("inmemory"));
+	if (!RTEST(val))
```
vmg: Doesn't look right :)

arthurschreiber: I think `RTEST` returns `1` for truthy values and `0` for falsy ones, so you could change this (and similar code) to just: `ret->inmemory = RTEST(rb_hash_aref(rb_options, CSTR2SYM("inmemory")));`.

carlosmn: yeah, it was a copy-paste error, wrote some tests and they failed and now it's fixed.

carlosmn:
I've squashed everything together, and this should be all that we need API-wise to perform rebases, in-memory or not.

I'm writing docs now, but I would still like some guidance in how to deal with `#commit`. Converting from C, the first and third args are optional. This is because we generally ask for the author first and then the committer. But in this case, you can just leave the author as `NULL`. This makes it look weird in ruby. I suppose the message we can just leave as an optional last arg, but I'm not sure what to do about the author. Should we accept a hash with `:author` and `:message` so you can specify either or both of them? Leave as-is?

vmg:
@carlosmn: Use named arguments, they're supported in new Rubies and will fix the issue alright. I think we have several examples of named argument parsing already.

carlosmn:
Does ruby implement this by passing a hash so I pass just `":"` to `rb_scan_args()` or is there something more to it?

PR #575, opened 2016-02-15
Added Rugged::Walker#count

YorickPeterse:
This method counts the amount of objects a Walker would iterate over
without allocating any Ruby objects. If one were to use the default
"count" method (as added by Enumerable) every call would end up
allocating Rugged::Commit objects for every commit (as "count" simply
calls "each").

Timing wise the differences aren't _that_ big. For example, take the
following benchmark:

    require 'rugged'
    require 'benchmark/ips'

    rugged = Rugged::Repository.new('path/to/gitlab-ce.git')
    from   = rugged.rev_parse_oid('5-4-stable')
    to     = rugged.rev_parse_oid('master')

    Benchmark.ips do |bench|
      bench.report 'each' do
        walker = Rugged::Walker.new(rugged)
        count  = 0

        walker.push(to)
        walker.hide(from)
        walker.each { count += 1 }
      end

      bench.report 'each_oid' do
        walker = Rugged::Walker.new(rugged)
        count  = 0

        walker.push(to)
        walker.hide(from)
        walker.each_oid { count += 1 }
      end

      bench.report 'count' do
        walker = Rugged::Walker.new(rugged)

        walker.push(to)
        walker.hide(from)
        walker.count
      end

      bench.compare!
    end

Here 'path/to/gitlab-ce.git' refers to a local clone of GitLab Community
Edition (https://gitlab.com/gitlab-org/gitlab-ce). This benchmark spits
out the following output:

    Warming up --------------------------------------
                    each     1.000  i/100ms
                each_oid     1.000  i/100ms
                   count     1.000  i/100ms
    Calculating -------------------------------------
                    each      6.753  (± 0.0%) i/s -     34.000
                each_oid      7.374  (± 0.0%) i/s -     37.000
                   count      7.557  (± 0.0%) i/s -     38.000

    Comparison:
                   count:        7.6 i/s
                each_oid:        7.4 i/s - 1.02x slower
                    each:        6.8 i/s - 1.12x slower

In other words, while the amount of object allocations can be greatly
reduced the actual runtime is not _that_ much better. Regardless, not
having to allocate as many objects as before is still a win.


ext/rugged/rugged_revwalk.c
```diff
@@ -192,6 +192,35 @@ static VALUE rb_git_walker_simplify_first_parent(VALUE self)
...
+ *    walker.count -> Fixnum
+ *
+ *  Returns the amount of objects a walker iterated over.
+ */
+static VALUE rb_git_walker_count(VALUE self)
+{
+    git_revwalk *walk;
+    git_oid commit_oid;
+    int error = 0;
+    long count = 0;
```
arthurschreiber: This probably can be a `uint64_t` as well.

ext/rugged/rugged_revwalk.c
```diff
@@ -192,6 +192,35 @@ static VALUE rb_git_walker_simplify_first_parent(VALUE self)
...
+    uint64_t limit = UINT64_MAX;
+
+    Data_Get_Struct(self, git_revwalk, walk);
+
+    while ((error = git_revwalk_next(&commit_oid, walk)) == 0) {
```
arthurschreiber: I believe if you do something like:

```
while ((error = git_revwalk_next(&commit_oid, walk)) == 0) && ++count != UINT64_MAX);
```

you could remove the limit variable altogether.

ext/rugged/rugged_revwalk.c
```diff
@@ -192,6 +192,35 @@ static VALUE rb_git_walker_simplify_first_parent(VALUE self)
...
+        count++;
+
+        if (--limit == 0)
+            break;
+    }
+
+    if (error != GIT_ITEROVER)
+        rugged_exception_check(error);
+
+    return INT2NUM(count);
```
arthurschreiber: If `count` is changed to `uint64_t`, you need to change this to `ULONG2NUM`.

ext/rugged/rugged_revwalk.c
YorickPeterse: I left it as `long` since I'm not 100% sure if `NUM2INT` can deal with `uint64_t` but I'll give it a try.

arthurschreiber:
LGTM, except for a few style suggestions.

YorickPeterse:
@arthurschreiber I adjusted the code based on your suggestions.

arthurschreiber:
@YorickPeterse Thanks. One last thing - can you change spaces to tabs? :heart:

arthurschreiber:
(in the C code only, that is)

YorickPeterse:
@arthurschreiber Oh shoot, I hadn't even noticed the file was using tabs. Sure.

YorickPeterse:
@arthurschreiber Done!

YorickPeterse:
@arthurschreiber Thanks!

arthurschreiber:
Hey @YorickPeterse,

I just checked the docs for `Enumerable#count`, and this change breaks the interface. `#count` can take either an optional parameter or take an optional block, and your implementation here won't work with either of those. Can you extend this to call the implementation coming from `Enumerable` by calling `rb_call_super` if a block or args are passed in? I know this is an edge case, but I'd prefer if we'd stick to the interface defined by `Enumerable`.

Thanks!

YorickPeterse:
@arthurschreiber Sure, I'll set up a new pull request for that.

PR #577, opened 2016-02-16
commit: raise when given bad data to extract the signature

carlosmn:
This lets the caller distinguish between a missing object, the wrong
kind of object and unsigned commits.

---

Limitations you discover when actually trying to integrate into the product.

### DIFF
--- a/ext/rugged/rugged_commit.c
+++ b/ext/rugged/rugged_commit.c
@@ -625,7 +625,7 @@ static VALUE rb_git_commit_header(VALUE self)
 static VALUE rb_git_commit_extract_signature(int argc, VALUE *argv, VALUE self)
 {
 	int error;
-	VALUE ret_arr;
+	VALUE ret;
 	git_oid commit_id;
 	const char *field;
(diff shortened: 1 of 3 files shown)


ext/rugged/rugged_commit.c
```diff
@@ -647,7 +648,8 @@ static VALUE rb_git_commit_extract_signature(int argc, VALUE *argv, VALUE self)
 		git_buf_free(&signed_data);
 	}
 
-	if (error == GIT_ENOTFOUND) {
+	err = giterr_last();
+	if (error == GIT_ENOTFOUND && err->klass == GITERR_OBJECT ) {
```
vmg: You can call `giterr_last()->klass` here, which probably looks cleaner than the extra `err` variable.

carlosmn: Yep, originally I was going to need it twice, but we now guarantee the error gets set. Changed.

vmg:
Looks great.

PR #580, opened 2016-03-01
Strict object creation

ethomson:
This updates libgit2 to [`edaffe2`](https://github.com/libgit2/libgit2/commit/edaffe22a205c57022cc365ab20dcbf7e22f68c4), which turns on stricter object validation for functions like the treebuilder.  Additionally, add a `Tree::Builder` test that ensures that we throw when adding a new tree entry that is nonexistent, but that we can disable this behavior with `Rugged::Settings['strict_object_creation']`.

/cc @vmg

### DIFF
--- a/ext/rugged/rugged_settings.c
+++ b/ext/rugged/rugged_settings.c
@@ -98,6 +98,11 @@ static VALUE rb_git_set_option(VALUE self, VALUE option, VALUE value)
 		set_search_path(GIT_CONFIG_LEVEL_SYSTEM, value);
 	}
 
+	else if (strcmp(opt, "strict_object_creation") == 0) {
+		int strict = RTEST(value) ? 1 : 0;
+		git_libgit2_opts(GIT_OPT_ENABLE_STRICT_OBJECT_CREATION, strict);
+	}
+
 	else {
 		rb_raise(rb_eArgError, "Unknown option specified");
 	}
--- a/test/tree_test.rb
+++ b/test/tree_test.rb
@@ -143,4 +143,17 @@ class TreeUpdateTest < Rugged::TestCase
     assert_equal builder.remove("nonexistent file"), false
   end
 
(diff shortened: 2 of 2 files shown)


ext/rugged/rugged_settings.c
```diff
@@ -98,6 +98,11 @@ static VALUE rb_git_set_option(VALUE self, VALUE option, VALUE value)
 		set_search_path(GIT_CONFIG_LEVEL_SYSTEM, value);
 	}
 
+	else if (strcmp(opt, "strict_object_creation") == 0) {
+		int strict = (TYPE(value) == T_FALSE) ? 0 : 1;
```
ethomson: Much better; thanks!

carlosmn:
:+1:

test/tree_test.rb
```diff
@@ -143,4 +143,26 @@ def test_treebuilder_remove
...
+    end
+  end
+
+  def test_treebuilder_add_nonexistent_can_pass
+    Rugged::Settings['strict_object_creation'] = false
+    builder = Rugged::Tree::Builder.new(@repo, @repo.head.target.tree)
+    builder << { :type => :blob, :name => "another-readme", :oid => "0000000000000000000000000000000000000000", :filemode => 0100644 }
+    newtree = builder.write
+    assert_equal "7c98360ac03064bb67c6f0949e6a354155ce1b04", newtree
+    Rugged::Settings['strict_object_creation'] = true
```
vmg: Putting this inside an "ensure" block would be nice.

ethomson: Why yes, it certainly would.  Fixed.

vmg: Ruby pro-tip: function definitions are implicit blocks, so you don't need a `begin` to place an `ensure` before the end. :ok_hand:

PR #582, opened 2016-03-04
Release 0.24.0

ethomson:
We've been incrementing 0.24.0 as a beta since [way back in October](https://github.com/libgit2/rugged/commit/fb17354c).  Now that libgit2 is [updating to 0.24](https://github.com/libgit2/libgit2/issues/3626) themselves, update to match.


vmg:
Yeeeey!

PR #583, opened 2016-03-09
ext: Use `xmalloc` instead of `malloc`

ethomson:
Use `xmalloc` instead of `malloc`, as `xmalloc` will fail gracefully, raising an exception and [returning to Ruby itself](https://github.com/ruby/ruby/blob/5fa5b50e587918f7003a9a6579e655896e82aa2d/eval.c#L589) while `malloc` will merrily carry-on and we will doubtless try to dereference a null pointer.


arthurschreiber:
:heart:

PR #584, opened 2016-03-09
The new Expand OIDs API

vmg:
As implemented by the almighty @ethomson :ok_hand:

### DIFF
--- a/ext/rugged/rugged_repo.c
+++ b/ext/rugged/rugged_repo.c
@@ -1045,10 +1045,13 @@ static VALUE rb_git_repo_read_header(VALUE self, VALUE hex)
 /**
  *  call-seq:
  *    repo.expand_oids([oid..], object_type = :any) -> hash
+ *    repo.expand_oids([oid..], object_type = [type..]) -> hash
  *
  *  Expand a list of short oids to their full value, assuming they exist
- *  in the repository. If `object_type` is passed, OIDs are expected to be
- *  of the given type.
+ *  in the repository. If `object_type` is passed and is an array, it must
+ *  be the same length as the OIDs array. If it's a single type name, all
+ *  OIDs will be expected to resolve to that object type. OIDs that don't
+ *  match the expected object types will not be expanded.
  *
  *  Returns a hash of `{ short_oid => full_oid }` for the short OIDs which
  *  exist in the repository and match the expected object type. Missing OIDs
--- a/ext/rugged/rugged_repo.c
+++ b/ext/rugged/rugged_repo.c
@@ -1058,59 +1058,64 @@ static VALUE rb_git_repo_expand_oids(int argc, VALUE *argv, VALUE self)
 {
 	VALUE rb_result, rb_oids, rb_expected_type;
 
-	git_otype expected_type = GIT_OBJ_ANY;
-
 	git_repository *repo;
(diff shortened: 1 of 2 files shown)


ext/rugged/rugged_repo.c
```diff
@@ -1058,59 +1058,64 @@ static VALUE rb_git_repo_expand_oids(int argc, VALUE *argv, VALUE self)
...
+	int error;
 
 	Data_Get_Struct(self, git_repository, repo);
-
 	rb_scan_args(argc, argv, "11", &rb_oids, &rb_expected_type);
 
 	Check_Type(rb_oids, T_ARRAY);
-	expected_type = rugged_otype_get(rb_expected_type);
+	expand_count = RARRAY_LEN(rb_oids);
+	expand = alloca(expand_count * sizeof(git_odb_expand_id));
```
vmg: Come at me bro

ext/rugged/rugged_repo.c
```diff
@@ -1058,59 +1058,64 @@ static VALUE rb_git_repo_expand_oids(int argc, VALUE *argv, VALUE self)
...
+		rugged_exception_check(
+			git_oid_fromstrn(&expand[i].id, RSTRING_PTR(rb_hex), RSTRING_LEN(rb_hex))
+		);
+		expand[i].length = RSTRING_LEN(rb_hex);
+	}
 
-	for (i = 0; i < RARRAY_LEN(rb_oids); ++i) {
-		VALUE hex_oid = rb_ary_entry(rb_oids, i);
-		git_oid found_oid;
+	if (TYPE(rb_expected_type) == T_ARRAY) {
```
ethomson: Wow aren't you fancy.  Do you want to update the documentation to reflect that you can now pass this fancy new array of types?

vmg: Good idea!

ext/rugged/rugged_repo.c
carlosmn: Would `xalloca()` be more appropriate here so we don't crash as badly when someone passes a million entries?

vmg: https://github.com/ruby/ruby/blob/75b61aa31646a2b598a5a64f3eeacdd0799e51ff/regint.h#L244

History lesson: MRI maps a magical page at the bottom (top) of the stack, so that accessing it raises a trap that can be captured and re-raised as a `StackOverflowError` in Ruby-land. It *needs* to do this because, well, Ruby bytecode is interpreted in C, so a stack overflow in Ruby would mean an actual crash of the program as we smash into the VM's heap.

With this magical page in the stack, we're free to call something like `alloca` anywhere in a C extension; a potential stack overflow in our C code will be handled by MRI in the same way it would handle a stack overflow in our Ruby code. And that's why `xalloca` is simply an alias for `alloca`. And that's also why I can write this code and sleep at night just fine.

:ok_hand:

arthurschreiber: If someone passes in a lot of entries, and you later fail the `Check_Type(rb_hex, T_STRING);` on the last entry, this is going to leak a looooot of memory. You'll want to check the type (and do all other things that can cause an exception) before you call `xalloca`/``alloca`.

PR #585, opened 2016-03-15
Add a way to create a commit as a buffer

carlosmn:
This is instead of writing it out to disk.

### DIFF
--- a/ext/rugged/rugged_commit.c
+++ b/ext/rugged/rugged_commit.c
@@ -693,11 +693,77 @@ static VALUE rb_git_commit_extract_signature(int argc, VALUE *argv, VALUE self)
 	return ret;
 }
 
+/*
+ *  call-seq:
+ *    Commit.create_to_s(repository, data = {}) -> str
+ *
+ *  Create a string with the contents of the commit, created with the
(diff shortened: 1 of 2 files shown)


ext/rugged/rugged_commit.c
```diff
@@ -662,11 +693,77 @@ static VALUE rb_git_commit_extract_signature(int argc, VALUE *argv, VALUE self)
 	return ret;
 }
 
+/*
+ *  call-seq:
+ *    Commit.create_buffer(repository, data = {}) -> commit
```
vmg: This should say `-> String`. But maybe `create_buffer` is not the ideal name here. `Commit#compose` maybe?

carlosmn: `#compose` seems like a weird name... *we're* not composing anything here... maybe `#create_as_string()` or something with `in_memorry` in the name?

vmg: `#create_to_s` sounds more Ruby-like

carlosmn: Let's do that then.

carlosmn:
Updated with the new name.

PR #589, opened 2016-03-29
extconf: Use MSYS Makefiles with CMake on Windows

sschuberth:
On Windows, Ruby-DevKit is MSYS based. As such we should use MSYS
Makefiles with CMake. Otherwise it might happen that CMake picks a
gmake.exe from somewhere instead of make.exe from Ruby-DevKit as CMake
seems to prefer gmake over make.

In a concrete case, when doing "gem install rugged" on Windows with
Strawberry Perl installed and in PATH, CMake was picking

    C:/Strawberry/c/bin/gmake.exe

instead of

    C:/Ruby-DevKit/bin/make.exe

While the latter is linked against MSYS-1.0.DLL and understands Unix-style
paths on Windows, the first is a native Windows executable linked against
MSVCRT.DLL that does not. This resulted in path mangling issues leading to
CMake's "Detecting C compiler ABI info" step to fail because the helper
executable failed to build, which in turn led to CMAKE_SIZEOF_VOID_P not
being set and libgit2 being unable to detect the architecture.

### DIFF
--- a/ext/rugged/extconf.rb
+++ b/ext/rugged/extconf.rb
@@ -67,7 +67,9 @@ else
     Dir.mkdir("build") if !Dir.exists?("build")
 
     Dir.chdir("build") do
-      sys("cmake .. -DBUILD_CLAR=OFF -DTHREADSAFE=ON -DBUILD_SHARED_LIBS=OFF -DCMAKE_C_FLAGS=-fPIC -DCMAKE_BUILD_TYPE=RelWithDebInfo -G \"Unix Makefiles\"")
+      # On Windows, Ruby-DevKit is MSYS-based, so ensure to use MSYS Makefiles.
+      generator = "-G \"MSYS Makefiles\"" if windows?
+      sys("cmake .. -DBUILD_CLAR=OFF -DTHREADSAFE=ON -DBUILD_SHARED_LIBS=OFF -DCMAKE_C_FLAGS=-fPIC -DCMAKE_BUILD_TYPE=RelWithDebInfo #{generator}")
       sys(MAKE)
 
       # "normal" libraries (and libgit2 builds) get all these when they build but we're doing it


sschuberth:
@ethomson This should fix the issues around `CMAKE_SIZEOF_VOID_P` I had with libgit2 (https://github.com/libgit2/libgit2/pull/3526 and https://github.com/libgit2/libgit2/pull/3708).

ethomson:
Hmmm.  So I think this is probably okay - using msys makefiles here probably makes more sense than hardcoding unix.  I'm a bit curious if we need to hardcode the generator at all, though.

I also wonder if the right long-term solution here isn't to [embed the DLLs in the gem](http://tenderlovemaking.com/2009/05/07/fat-binary-gems-make-the-rockin-world-go-round.html).  Relying on people to have Ruby-DevKit just to install a gem is sort of a bummer.

But that's a more aggressive change, of course.

sschuberth:
> I'm a bit curious if we need to hardcode the generator at all, though.

I think we do. On my Windows machine, which also has Visual Studio installed, CMake would otherwise use the Visual Studio generator when omitting `-G` even though `/C/Ruby-DevKit/mingw/bin/gcc.exe` is in `PATH`.

>  Relying on people to have Ruby-DevKit just to install a gem is sort of a bummer.

I absolutely agree, but like you say, that's a bigger change that I'd not like to do right now.

arthurschreiber:
@ethomson The reason we don't publish a binary gem for windows is that the rugged tests fail when running on windows. This _can_ be fixed, and I think there's a branch I published that tries to do that, but I don't think I got around to finishing that (I think I should be able to pick this up again very soon™).

@sschuberth The change itself looks good. As you pointed out, I hardcoded this to use MINGW Makefiles because CMake tries to generate VS Makefiles instead, which won't work with the Ruby DevKit, and MINGW worked for both. I think it would be better if we only specify the makefile format on windows, and keep the default on all other platforms.

sschuberth:
@arthurschreiber I've updated the PR, please have another look.

sschuberth:
Thanks for the merge! Any chance to release a new Gem soonish with this fix?

PR #591, opened 2016-04-21
Extend Rebase

arthurschreiber:
This extends `Rugged::Rebase.new` to correctly handle `Rugged::Reference` and `Rugged::Commit` objects. Passing Strings still works, but they're now treated as full revspecs, whereas previously they've only been treated as refnames.

This also takes a stab at fixing some memory leak issues.

### DIFF
--- a/ext/rugged/rugged_rebase.c
+++ b/ext/rugged/rugged_rebase.c
@@ -27,6 +27,8 @@
 extern VALUE rb_mRugged;
 extern VALUE rb_cRuggedIndex;
 extern VALUE rb_cRuggedRepo;
+extern VALUE rb_cRuggedCommit;
+extern VALUE rb_cRuggedReference;
 
 VALUE rb_cRuggedRebase;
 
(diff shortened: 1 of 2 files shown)


arthurschreiber:
2 test cases are currently failing because of a bug in `git_rebase_abort`, which fails when aborting a rebase that was initialized using annotated commits looked up through a revspec. 😢

ext/rugged/rugged_rebase.c
```diff
@@ -67,6 +69,63 @@ VALUE rugged_rebase_new(VALUE klass, VALUE owner, git_rebase *rebase)
 	return rb_rebase;
 }
 
+struct get_annotated_commit_args {
```
arthurschreiber: @carlosmn Do you think we could convert this pattern (lines 72 up to 127) into a macro?

arthurschreiber:
@ethomson It looks like your fixes in libgit2/libgit2#3748 didn't catch the case of aborting a rebase that was started using an annotated commit that was looked up using a shortened commit sha. 😢

carlosmn:
This issue stems from using `_from_revspec()` since it'll store whatever went in as the branch name, which will make the rebase think we're dealing with branches even though we could have stuff like `heads/master@{2.days.ago}` as the branch.

ethomson:
> This issue stems from using _from_revspec()

Right.  `from_revspec` is conflating reference names and description.  I'm unpacking this mess now.

ethomson:
(More accurately, `rebase` suddenly leapt up and decided that the description was meaningful.)

ethomson:
Should be fixed in https://github.com/libgit2/libgit2/pull/3758.

Note, though...  this is getting a little edge casey.  If you're passing in strange `branch` values (like things that are revparsed) then you really should be doing this from a `inmemory` rebase.  You wouldn't ever do this with a working directory.

That's not to say that this wasn't a bug in libgit2 (it obviously was) but that maybe the test should reflect that.

arthurschreiber:
Merged this as part of #597

PR #594, opened 2016-04-26
Upgrade to the latest libgit2 master

arthurschreiber:
/cc @carlosmn

### DIFF
--- a/ext/rugged/rugged_blob.c
+++ b/ext/rugged/rugged_blob.c
@@ -231,30 +231,6 @@ static VALUE rb_read_check(VALUE pointer) {
 	return rb_buffer;
 }
 
-static int cb_blob__get__chunk(char *content, size_t max_length, void *data)
-{
-	VALUE rb_buffer, rb_args[2];
-	size_t str_len, safe_len;
-	struct rugged_cb_payload *payload = data;
-
-	rb_args[0] = payload->rb_data;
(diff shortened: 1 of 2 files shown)


ext/rugged/rugged_blob.c
```diff
@@ -300,18 +275,34 @@ static VALUE rb_git_blob_from_io(int argc, VALUE *argv, VALUE klass)
...
+	rb_read_args[0] = rb_io;
+	rb_read_args[1] = INT2FIX(max_length);
+
+	do {
+		rb_buffer = rb_protect(rb_read_check, (VALUE)rb_read_args, &exception);
+
+		if (exception)
+			goto cleanup;
+
+		if (NIL_P(rb_buffer))
```
vmg: You need a `TYPE(rb_buffer)` check here. Otherwise this will segfault if our callback returns something that is not a string.

arthurschreiber: You're right. I think I'll put that check into the ´rb_read_check´ to make sure we don't leak memory in that case.

vmg: Even better 👍

vmg: Actually, that check is _already_ in `rb_read_check`. Nevermind, we're all good. 😅

arthurschreiber: Ha! 😆

PR #600, opened 2016-05-06
Rebase: Better error handling in case of empty patches.

arthurschreiber:
Right now, Rugged will throw a `Rugged::RebaseError` when trying to commit an empty rebase patch. This error type can also be thrown for a bunch of other error conditions, which leaves consumers testing for the error's `message` to distinguish between these different error types.

Let's introduce a new exception class instead.

/cc @vmg @ethomson @carlosmn

### DIFF
--- a/ext/rugged/rugged_rebase.c
+++ b/ext/rugged/rugged_rebase.c
@@ -189,11 +189,10 @@ cleanup:
 	git_annotated_commit_free(upstream);
 	git_annotated_commit_free(onto);
 
-	if (error) {
-		rugged_exception_check(error);
-	} else if (exception) {
+	if (exception)
 		rb_jump_tag(exception);
-	}
+
+	rugged_exception_check(error);
 
 	return rugged_rebase_new(klass, rb_repo, rebase);
 }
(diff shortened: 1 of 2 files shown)


carlosmn:
This definitely should be an error you can distinguish, although I wonder if an exception is the best way to do this. Can you not continue?

arthurschreiber:
> Can you not continue?

We could on the C side, but that would just hide the problem. I think it's actually interesting for consumers of the library to know that this has happened and e.g. leave this to the actual user to decide what to do.

carlosmn:
>We could on the C side, but that would just hide the problem.

I don't mean we should silently continue, but an exception seems like the wrong way to about it since it's not really an error, nor is it particularly unexpected. I'm thinking of the use-case where you're doing a `git-rebase` and showing the progress on a screen (terminal/GUI/a website). A patch that doesn't end up adding changes would just be one situation to consider for the output, and you I don't think you'd ever actually want to halt the rebase because of it.

vmg:
@arthurschreiber: What @carlosmn suggests is a return value instead of an Exception, since Exceptions (even in an exception-happy language like Ruby) are meant for exceptional events. Applying an empty patch is not an error. So ideally we'd come up with a way to display this information through the API without having to `catch` anything.

arthurschreiber:
That sounds reasonable. What do you think about returning `nil` in those cases? I think that would be sensible: If there were changes, and a commit was created, we return the oid. If there are no changes, no new commit can be created, and we return `nil`.

vmg:
👌

PR #602, opened 2016-05-10
Tag oids as US_ASCII

tenderlove:
We should be tagging oids as US-ASCII (or UTF-8).  We know the encoding
of the bytes in that string, so we should be indicating that encoding to
clients of the library.

### DIFF
--- a/ext/rugged/rugged.h
+++ b/ext/rugged/rugged.h
@@ -176,7 +176,7 @@ static inline VALUE rugged_create_oid(const git_oid *oid)
 {
 	char out[40];
 	git_oid_fmt(out, oid);
-	return rb_str_new(out, 40);
+	return rb_usascii_str_new(out, 40);
 }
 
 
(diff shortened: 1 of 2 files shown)


ethomson:
Out of curiosity, is there an advantage to tag these as one over the other?

arthurschreiber:
@ethomson I don't think there's a huge difference between tagging this one way or the other. Ruby knows that ascii is a subset of UTF-8, so merging UTF-8 and ascii strings is pretty straightforward and should just work.

tenderlove:
@ethomson as @arthurschreiber says, there's really no difference.  The only reason I even thought of it is because I noticed that methods like `Reference#target_id` will [return either `ASCII-8BIT` or `UTF-8`](https://github.com/libgit2/rugged/blob/abaf705fe177c86bb94afee957b763db2a156f6f/ext/rugged/rugged_reference.c#L177-L181).  This patch makes them return `US-ASCII` or `UTF-8`.  It's nice to have methods always return the same thing, but I think as long as they're both tagged correctly, it doesn't matter.

PR #605, opened 2016-05-18
Bump libgit2, include tree updater

carlosmn:
Bump libgit2 now that we trust it a bit more, and add the more efficient tree updater method.

I've tried to make the docs clear, but I dunno, I'd appreciate pointers if it can be made better.

### DIFF
--- a/ext/rugged/rugged_tree.c
+++ b/ext/rugged/rugged_tree.c
@@ -702,6 +702,31 @@ static VALUE rb_git_tree_merge(int argc, VALUE *argv, VALUE self)
 	return rugged_index_new(rb_cRuggedIndex, rb_repo, index);
 }
 
+static git_oid empty_tree = {{ 0x4b, 0x82, 0x5d, 0xc6, 0x42, 0xcb, 0x6e, 0xb9, 0xa0, 0x60,
+			       0xe5, 0x4b, 0xf8, 0xd6, 0x92, 0x88, 0xfb, 0xee, 0x49, 0x04 }};
+
+/*
+ *  call-seq:
+ *    Tree.empty(repo) -> tree
+ *
+ *  Look up the empty tree in the given repository +repo+. The empty
+ *  tree's id is hard-coded to exist in a repository.
+ *
+ *  Returns a new instance of the empty tree.
+ */
+static VALUE rb_git_tree_empty(VALUE self, VALUE rb_repo)
+{
+	git_repository *repo;
+	git_tree *tree;
+
+	rugged_check_repo(rb_repo);
+	Data_Get_Struct(rb_repo, git_repository, repo);
+
+	rugged_exception_check(git_tree_lookup(&tree, repo, &empty_tree));
+
+	return rugged_object_new(rb_repo, (git_object *) tree);
+}
+
 /**
  * Parse the updates and convert them into libgit2 ones. They will be
  * heap-allocated and returned in 'out'. The strings will also be
@@ -765,14 +790,10 @@ on_error:
 
 /*
  *  call-seq:
- *    Tree.create_updated(repository, baseline, updates)
+ *    tree.update(updates)
  *
- *  Create a new Rugged::Tree based on +baseline+ and applying the
+ *  Create a new Rugged::Tree based on the curent one by applying the
  *  changes described in +updates+.
- *
- *  If an optional +baseline+ is given, the returned Tree will be
- *  based on that tree, otherwise the starting point will be the empty
- *  tree.
  *
  *  The updates are given as a list of +Hash+ containing:
  *
@@ -790,7 +811,7 @@ on_error:
  *    intermediate trees will be created.
  *
  */
-static VALUE rb_git_tree_create_updated(VALUE self, VALUE rb_repo, VALUE rb_tree, VALUE rb_updates)
+static VALUE rb_git_tree_update(VALUE self, VALUE rb_updates)
 {
 	git_repository *repo;
 	git_tree *tree = NULL;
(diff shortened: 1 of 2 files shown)


ext/rugged/rugged_tree.c
```diff
@@ -702,6 +702,125 @@ static VALUE rb_git_tree_merge(int argc, VALUE *argv, VALUE self)
...
+ *
+ *  :filemode ::
+ *    The octal filemode for the entry. This is ignored for remvals.
+ *
+ *  :path ::
+ *    The path of the entry. This may contain slashes and the
+ *    intermediate trees will be created.
+ *
+ */
+static VALUE rb_git_tree_create_updated(VALUE self, VALUE rb_repo, VALUE rb_tree, VALUE rb_updates)
```
arthurschreiber: What do you think about making this an instance method instead of a class method?

carlosmn: That would work when you do have a baseline, but if you want to create a new tree without a base, what would you do?

carlosmn: This is the same reason we take both a repo and a tree. When there is no tree to start from, we don't know which repo we should be working with.

carlosmn: @arthurschreiber How 'bout this: We have _both_ and in the class method you pass in the repo, in the instance method we assume the baseline is `self` and use the repo from there.

I'm not sure if that's going to be more confusing, though.

arthurschreiber: What do you think about making `Tree.new(repo)` return the empty tree object? Then you could do `Tree.new(repo).update(...)` and have the interface unified.

carlosmn: That seems like a really weird way of looking up the empty tree. We do already return an empty tree if you look up the id, so maybe maybe have `Tree.empty(repo)` do what your `Tree.new(repo)` would do and then we can still unify the interface.

carlosmn: You can already do `Tree.lookup(repo, <empty tree id>)` but I certainly don't remember what the id is off-hand. Maybe some globals would help there. `Tree.lookup(repo, EMPTY_TREE_OID).update(...)`?

ext/rugged/rugged_tree.c
```diff
@@ -702,6 +702,139 @@ static VALUE rb_git_tree_merge(int argc, VALUE *argv, VALUE self)
...
+ *  Create a new Rugged::Tree based on the curent one by applying the
+ *  changes described in +updates+.
+ *
+ *  The updates are given as a list of +Hash+ containing:
+ *
+ *  :action ::
+ *    +:upsert+ or +:remove+ to add/insert an entry, or to remove it resp.
+ *
+ *  :oid ::
+ *    The +oid+ of the entry. This is ignored for removals.
```
vmg: > This is ignored for removals.

This is actually not true. If you try to add a removal entry to the list of operations without an OID or a filemode, Rugged will raise an exception, even though it doesn't really use the value of the OID or the filemode. I would suggest fixing that and adding a removal test -- all the existing test cases are for upserts.

ext/rugged/rugged_tree.c
```diff
@@ -702,6 +702,139 @@ static VALUE rb_git_tree_merge(int argc, VALUE *argv, VALUE self)
...
+
+		rb_action   = rb_hash_aref(rb_update, CSTR2SYM("action"));
+		rb_oid      = rb_hash_aref(rb_update, CSTR2SYM("oid"));
+		rb_filemode = rb_hash_aref(rb_update, CSTR2SYM("filemode"));
+		rb_path     = rb_hash_aref(rb_update, CSTR2SYM("path"));
+
+		if (!SYMBOL_P(rb_action) ||
+		    !RB_TYPE_P(rb_path, T_STRING) ||
+		    !RB_TYPE_P(rb_path, T_STRING) ||
+		    !RB_TYPE_P(rb_filemode, T_FIXNUM))
```
vmg: You need to move this check into the `upsert` branch.

ext/rugged/rugged_tree.c
```diff
@@ -702,6 +702,139 @@ static VALUE rb_git_tree_merge(int argc, VALUE *argv, VALUE self)
...
+			goto on_error;
+
+		rb_action   = rb_hash_aref(rb_update, CSTR2SYM("action"));
+		rb_oid      = rb_hash_aref(rb_update, CSTR2SYM("oid"));
+		rb_filemode = rb_hash_aref(rb_update, CSTR2SYM("filemode"));
+		rb_path     = rb_hash_aref(rb_update, CSTR2SYM("path"));
+
+		if (!SYMBOL_P(rb_action) ||
+		    !RB_TYPE_P(rb_path, T_STRING) ||
+		    !RB_TYPE_P(rb_path, T_STRING) ||
```
vmg: This check is duplicated. I think you wanted to write `RB_TYPE_P(rb_oid, T_STRING)`. But that would also be wrong, the check needs to be moved into the `upsert` branch. :)

carlosmn:
Fixed the various checks to check only when we need to do it. The added removal check has surfaced a bug in libgit2 when removing a deeper file.

ext/rugged/rugged_tree.c
arthurschreiber: I wonder if we shouldn't just move `Tree.empty` to be implemented in Ruby. We could also have a constant `Rugged::Tree::EMPTY_OID` or something like that to expose the OID of an empty tree.

carlosmn: I suppose we could, but everything else in already here.

ext/rugged/rugged_tree.c
```diff
@@ -702,6 +702,139 @@ static VALUE rb_git_tree_merge(int argc, VALUE *argv, VALUE self)
...
+	git_tree *tree = NULL;
+	git_tree_update *updates;
+	int nupdates, error;
+	git_oid id;
+
+	Data_Get_Struct(self, git_tree, tree);
+	repo = git_tree_owner(tree);
+
+	if (NIL_P(rb_updates))
+		rb_raise(rb_eTypeError, "The list of upates cannot be nil");
```
vmg: This is superfluous, we are already doing a `Check_Type` in `parse_tree_updates`.

vmg:
> The added removal check has surfaced a bug in libgit2 when removing a deeper file.

Good :D

PR #619, opened 2016-07-08
Tag strings without a known encoding as ASCII-8BIT.

arthurschreiber:
Almost all of the string used inside a git repository are encoding unaware.

This includes things like refnames, commit metadata, path names, and a lot more. The exception is commit messages, which can optionally be tagged with an encoding through a header in the commit metadata. We should only tag strings with an encoding if we know the exact encoding, and otherwise tag them as ASCII 8-BIT (binary).

### DIFF
--- a/ext/rugged/rugged_commit.c
+++ b/ext/rugged/rugged_commit.c
@@ -47,7 +47,7 @@ VALUE rb_cRuggedCommit;
 static VALUE rb_git_commit_message_GET(VALUE self)
 {
 	git_commit *commit;
-	rb_encoding *encoding = rb_utf8_encoding();
+	rb_encoding *encoding = rb_ascii8bit_encoding();
 	const char *encoding_name;
 	const char *message;
 
@@ -82,9 +82,7 @@ static VALUE rb_git_commit_committer_GET(VALUE self)
 	git_commit *commit;
 	Data_Get_Struct(self, git_commit, commit);
 
-	return rugged_signature_new(
-		git_commit_committer(commit),
-		git_commit_message_encoding(commit));
+	return rugged_signature_new(git_commit_committer(commit));
 }
 
 /*
@@ -107,9 +105,7 @@ static VALUE rb_git_commit_author_GET(VALUE self)
 	git_commit *commit;
 	Data_Get_Struct(self, git_commit, commit);
 
-	return rugged_signature_new(
-		git_commit_author(commit),
-		git_commit_message_encoding(commit));
+	return rugged_signature_new(git_commit_author(commit));
 }
 
 /*
@@ -577,7 +573,7 @@ static VALUE rb_git_commit_to_mbox(int argc, VALUE *argv, VALUE self)
 
 	if (error) goto cleanup;
 
-	rb_email_patch = rb_enc_str_new(email_patch.ptr, email_patch.size, rb_utf8_encoding());
+	rb_email_patch = rb_str_new(email_patch.ptr, email_patch.size);
 
 	cleanup:
 
@@ -598,9 +594,6 @@ static VALUE rb_git_commit_header_field(VALUE self, VALUE rb_field)
 {
 	git_buf header_field = { 0 };
 	git_commit *commit = NULL;
-
-	const char *encoding_name;
-	rb_encoding *encoding = rb_utf8_encoding();
 	VALUE rb_result;
 
 	int error;
@@ -617,11 +610,7 @@ static VALUE rb_git_commit_header_field(VALUE self, VALUE rb_field)
 		rugged_exception_check(error);
 	}
 
-	encoding_name = git_commit_message_encoding(commit);
-	if (encoding_name != NULL)
-		encoding = rb_enc_find(encoding_name);
-
-	rb_result = rb_enc_str_new(header_field.ptr, header_field.size, encoding);
+	rb_result = rb_str_new(header_field.ptr, header_field.size);
 	git_buf_free(&header_field);
 	return rb_result;
 }
@@ -640,7 +629,7 @@ static VALUE rb_git_commit_header(VALUE self)
 	Data_Get_Struct(self, git_commit, commit);
 
 	raw_header = git_commit_raw_header(commit);
-	return rb_str_new_utf8(raw_header);
+	return rb_str_new2(raw_header);
 }
 
 /*
@@ -753,7 +742,7 @@ cleanup:
 
 	rugged_exception_check(error);
 
-	ret = rb_str_new_utf8(buf.ptr);
+	ret = rb_str_new(buf.ptr, buf.size);
 	git_buf_free(&buf);
 
 	return ret;
--- a/ext/rugged/rugged_submodule.c
+++ b/ext/rugged/rugged_submodule.c
@@ -535,7 +535,7 @@ static VALUE rb_git_submodule_name(VALUE self)
 
 	name = git_submodule_name(submodule);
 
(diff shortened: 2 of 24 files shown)


ext/rugged/rugged_commit.c
```diff
@@ -107,9 +105,7 @@ static VALUE rb_git_commit_author_GET(VALUE self)
 	git_commit *commit;
 	Data_Get_Struct(self, git_commit, commit);
 
-	return rugged_signature_new(
-		git_commit_author(commit),
-		git_commit_message_encoding(commit));
+	return rugged_signature_new(git_commit_author(commit));
```
carlosmn: This is its own can of worms, but for some use-cases we have assumed that the encoding of the signature is the same as of the commit message. This sometimes even holds true, and this is a place where we kinda have a good idea of what the encoding is likely to be, rather than no idea like in the other cases.

tenderlove: I think this is probably an OK assumption to make.  I haven't seen any exceptions regarding this in production (only the path stuff)

ext/rugged/rugged_submodule.c
```diff
@@ -554,7 +554,7 @@ static VALUE rb_git_submodule_url(VALUE self)
 
 	url = git_submodule_url(submodule);
 
-	return url ? rb_str_new_utf8(url) : Qnil;
+	return url ? rb_str_new2(url) : Qnil;
```
tenderlove: I think URLs should be US-ASCII.

ext/rugged/rugged_tree.c
```diff
@@ -44,7 +44,7 @@ static VALUE rb_git_treeentry_fromC(const git_tree_entry *entry)
 
 	rb_entry = rb_hash_new();
 
-	rb_hash_aset(rb_entry, CSTR2SYM("name"), rb_str_new_utf8(git_tree_entry_name(entry)));
+	rb_hash_aset(rb_entry, CSTR2SYM("name"), rb_str_new2(git_tree_entry_name(entry)));
```
tenderlove: This is tangential to this PR, but I noticed that `git_tree_entry` has a length field for the name, but no accessor function.  Should we add an accessor function and use the length to create the string here?

test/branch_test.rb
```diff
@@ -116,8 +116,8 @@ def test_create_unicode_branch_nfc
     new_branch = @repo.create_branch(branch_name, "5b5b025afb0b4c913b4c338a42934a3863bf3644")
     refute_nil new_branch
 
-    assert_equal branch_name, new_branch.name
-    assert_equal "refs/heads/#{branch_name}", new_branch.canonical_name
+    assert_equal branch_name.force_encoding("ascii-8bit"), new_branch.name
+    assert_equal "refs/heads/#{branch_name}".force_encoding("ascii-8bit"), new_branch.canonical_name
```
tenderlove: Can you use `.b` rather than `force_encoding`?  I'm not sure if that exists on all Ruby versions we support (but it might).

test/reference_test.rb
```diff
@@ -179,8 +179,8 @@ def test_create_unicode_reference_nfc
     new_ref = @repo.references.create(ref_name, "5b5b025afb0b4c913b4c338a42934a3863bf3644")
     refute_nil new_ref
 
-    assert_equal ref_name, new_ref.name
-    assert_equal ref_name, new_ref.canonical_name
+    assert_equal ref_name.force_encoding("ascii-8bit"), new_ref.name
+    assert_equal ref_name.force_encoding("ascii-8bit"), new_ref.canonical_name
```
tenderlove: Could we just add `.b` on line 177 then eliminate these two?

tenderlove:
> US-ASCII 8-BIT

Just to be clear, US-ASCII and ASCII-8BIT are totally different.  US-ASCII is a known encoding, where ASCII-8BIT means "we don't know what the encoding of this string it".  All strings should have an encoding, ASCII-8BIT just means that we don't know what the encoding is.

tenderlove:
I was thinking about this last night.  I thought it might be nice if we could configure a repository with encodings.  Something like this:

```
repo = Rugged::Repository.new(ARGV[0], path_encoding: ::Encoding::UTF_8)
``

That way if you're dealing with a repository where you know the encoding of the paths, you can just configure the repo with the encoding to use.  We could provide options for the bits of data where we can't know the encoding like paths and tags (those are the only two I can think of, blobs should always be binary).

If we add this configuration value, then we can default it to UTF-8 in order to maintain backwards compatibility.

arthurschreiber:
> Just to be clear, US-ASCII and ASCII-8BIT are totally different.

Whoops, you're right, I mixed that up! 😄

arthurschreiber:
> I thought it might be nice if we could configure a repository with encodings.

I'm not sure this is a good solution. If you have different people that work on a repo, they might have set different encodings on their machines. I don't think that's too uncommon, especially for people working on windows. 😞

ethomson:
> If you have different people that work on a repo, they might have set different encodings on their machines. I don't think that's too uncommon, especially for people working on windows. 😞

ethomson:
> If you have different people that work on a repo, they might have set different encodings on their machines. I don't think that's too uncommon, especially for people working on windows. 😞

On Windows, at least, [NTFS stores filenames as UTF-16](https://msdn.microsoft.com/en-us/library/windows/desktop/dd317748%28v=vs.85%29.aspx) and there is, AFAIK, no way to use any other encoding.  This is fortunate, since the Windows Git implementations do an insane amount of UTF16 <-> UTF8 conversion to be able to talk to the Windows APIs, since they all speak UTF-16 (or UCS-2 in some cases.  Yay!)

Similarly, HFS+ will use a canonically decomposed UTF-16.

I'm just providing data here, I still think that returning these as `ASCII-8BIT` probably still makes a lot of sense.

ext/rugged/rugged_submodule.c
arthurschreiber: Are you sure URLs can't contain multibyte characters? What about IDNs?

ext/rugged/rugged_tree.c
arthurschreiber: I don't think tree entries can contain a null byte char in their name, but @carlosmn might know better.

arthurschreiber:
@tenderlove Is this going to break anything if we would push this into production as-is?

tenderlove:
I think it will work but we should test. I *think* we're retagging them as binary in all places so this should be OK

--
Aaron Patterson
http://tenderlovemaking.com/
I'm on an iPhone so I apologize for top posting.

ext/rugged/rugged_tree.c
carlosmn: It cannot. We have a field with the length so we don't keep recalculating it inside libgit2 but it's just a normal string.

ext/rugged/rugged_submodule.c
tenderlove: Pretty sure.  I think IDNs use [Punycode](https://en.wikipedia.org/wiki/Punycode).

Also:

qmk_firmware fish users aaron git qmk_firmware -fish 247x51 2019-02-15 08-11-20

tenderlove: (that said, I have no idea if git actually enforces it)

carlosmn: I think git would pass whatever garbage over to libcurl and have it figure it out.

scutuatua-crypto (approved):

PR #621, opened 2016-08-05
Fix Index#merge_file with custom labels

smcgivern:
Previously this required that the custom labels (for ours, theirs, and ancestor) were fixnums instead of strings, which contradicts the docs (and doesn't make a lot of sense :smiley: ).


vmg:
This was indeed wrong. Thank you!

PR #622, opened 2016-08-08
Add Rugged::Diff::Patch#bytesize

brianmario:
Similar to `Rugged::Diff::Patch#lines`, this returns the number of bytes in the patch text.

It accepts 3 optional boolean, keyword arguments:

* `include_context` - include context lines in the byte count
* `include_hunk_headers` - include hunk header lines in the byte count
* `include_file_headers` - include file header lines in the byte count

This needs inline-documentation written still but wanted to get this open for review early in case any changes to the API will mean only writing the docs once.


arthurschreiber:
Oh wow, this is looking great! I completely forgot about `git_patch_size`. ❤️

PR #623, opened 2016-08-12
Change Rugged::Patch#lines to accept count exclusion options

brianmario:
This is related to https://github.com/libgit2/rugged/pull/622 which added the `Rugged::Patch#bytesize` - which itself accepts options for what to include in the counts.

But, since `Rugged::Patch#lines` already existed and included the _total_ counts of context, addition and deletion lines - the options added to that method here are **exclusionary**. Meaning by default this method will continue to return the same value it always has, unless  you specify otherwise.

This feels annoyingly inconsistent with `Rugged::Patch#bytesize` but unless we're willing to break the API I'm not sure what else to do ;)

Thoughts?

### DIFF
--- a/ext/rugged/rugged_patch.c
+++ b/ext/rugged/rugged_patch.c
@@ -175,19 +175,56 @@ static VALUE rb_git_diff_patch_stat(VALUE self)
 
 /*
  *  call-seq:
- *    patch.lines -> int
+ *    patch.lines(options = {}) -> int
+ *
+ *  The following options can be passed in the +options+ Hash:
+ *
+ *  :exclude_context ::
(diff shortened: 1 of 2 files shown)


arthurschreiber:
This is really great, thanks for taking the time and implement this! 👍

I do think we should align the APIs. To not break too much, I think we should modify `#bytesize` to return the full diff size if no options are passed, and otherwise to support the same exclusion options as you added to `#lines` here.

I understand `#bytesize` and it's options were part of a beta release, so I understand your reluctance to change it, but that's what beta releases are for, right? 😉

brianmario:
@arthurschreiber done!

arthurschreiber:
Perfect! 😍

PR #626, opened 2016-08-19
Return GIT_PASSTHROUGH if credentials callback returns nill

### LINKED ISSUE #625: credentials callback not returning credentials causes a segmentation fault
```ruby
require 'rugged'

def credentials
  lambda do |_, _, _|
    return nil
  end
end

Rugged::Repository.clone_at 'https://github.com/example/my-project.git', 'my-project' , { credentials: credentials }
```

Calling `clone_at` with a credentials callback that returns nil against a repository that requires authentication causes a segmentation fault. When `on_headers_complete` tries to dereference `t->cred->credtype` (http.c:379) and `t->cred` is empty the segfault is triggered.

brandonweeks:
Fixed #625, see libgit2/libgit2#3902 for more information. Since triggering the error condition requires connecting to a repository that requires authentication I wasn't sure what a good way to test this would be.

### DIFF
--- a/ext/rugged/rugged_remote.c
+++ b/ext/rugged/rugged_remote.c
@@ -152,6 +152,9 @@ static int credentials_cb(
 
 	rb_protect(extract_cred, (VALUE)&args, &payload->exception);
 
+	if (!*cred)
+		return GIT_PASSTHROUGH;
+
 	return payload->exception ? GIT_ERROR : GIT_OK;
 }
 


carlosmn:
This doesn't look like the correct behaviour in this case. Returning `nil` from a credentials callback is a bug in that callback and it's completely different from indicating that the callback isn't set, including the error message which would now be misleading.

PR #627, opened 2016-08-25
Fix `Rugged::Patch#lines` and `Rugged::Patch#bytesize`

arthurschreiber:
I ran across an issue in `Rugged::Patch#lines` where the current implementation completely ignores the EOF newline markers, as it only sums up additions, deletions and context lines.

Also, it looks like @brianmario only applied documentation changes to `Rugged::Patch#bytesize` when aligning it with `Rugged::Patch#lines`, but did not modify the code. :trollface: (And I failed to see that during review).

This fixes both these issues. 👍


vmg:
This is really clean, Arthur. I love the flags approach. Good job here!

arthurschreiber:
Thanks @vmg 😍

PR #628, opened 2016-08-25
Add `Rugged::Patch#header`

arthurschreiber:
In some cases we might want to only generate the header of a `Rugged::Patch`, not the whole output.

### DIFF
--- a/ext/rugged/rugged_patch.c
+++ b/ext/rugged/rugged_patch.c
@@ -346,6 +346,22 @@ static int patch_print_cb(
 	return GIT_OK;
 }
 
+static int patch_print_header_cb(
+	const git_diff_delta *delta,
+	const git_diff_hunk *hunk,
+	const git_diff_line *line,
+	void *payload)
+{
+	VALUE rb_str = (VALUE)payload;
+
+	if (line->origin == GIT_DIFF_LINE_FILE_HDR) {
+		rb_str_cat(rb_str, line->content, line->content_len);
+	}
+
+	return GIT_OK;
+}
+
+
 /*
  *  call-seq:
  *    patch.to_s -> str
@@ -363,6 +379,24 @@ static VALUE rb_git_diff_patch_to_s(VALUE self)
 	return rb_str;
 }
 
+/*
+ *  call-seq:
+ *    patch.header -> str
+ *
+ *  Returns only the header of the patch as a string.
+ */
+static VALUE rb_git_diff_patch_header(VALUE self)
+{
+	git_patch *patch;
+	VALUE rb_str = rb_str_new(NULL, 0);
+	Data_Get_Struct(self, git_patch, patch);
+
+	rugged_exception_check(git_patch_print(patch, patch_print_header_cb, (void*)rb_str));
+
+	return rb_str;
+}
+
+
 void Init_rugged_patch(void)
 {
 	rb_cRuggedPatch = rb_define_class_under(rb_mRugged, "Patch", rb_cObject);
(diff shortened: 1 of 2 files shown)


ext/rugged/rugged_patch.c
```diff
@@ -332,16 +333,31 @@ static int patch_print_cb(
...
+	const git_diff_line *line,
+	void *payload)
+{
+	VALUE rb_buffer = (VALUE)payload;
+
+	if (line->origin == GIT_DIFF_LINE_FILE_HDR) {
+		rb_ary_push(rb_buffer, rb_str_new(line->content, line->content_len));
+	}
 
 	return GIT_OK;
```
brianmario: What do you think about returning `GIT_EUSER` (or I wonder if `GIT_ITEROVER` would work too?) once we hit the first line after the header?

That should stop the internal print loop from calling this callback once we have what we need. We'll just want to check the return value of `git_patch_print` before passing it to `rugged_exception_check` for this special error code.

arthurschreiber: Great idea! Just pushed another change for this! 👍

ext/rugged/rugged_patch.c
```diff
@@ -355,14 +372,35 @@ static int patch_print_cb(
 static VALUE rb_git_diff_patch_to_s(VALUE self)
 {
 	git_patch *patch;
-	VALUE rb_str = rb_str_new(NULL, 0);
+	VALUE rb_buffer = rb_ary_new();
+	Data_Get_Struct(self, git_patch, patch);
+
+	rugged_exception_check(git_patch_print(patch, patch_print_cb, (void*)rb_buffer));
+
+	return rb_ary_join(rb_buffer, Qnil);
```
brianmario: :+1: nice!

cc @tenderlove

brianmario:
Awesome! And good call on that little optimization in the `to_s` call.

I do wonder about the trade-off there with generating more "garbage" (all the small strings). Being as though each of those strings in the array will be a new allocation vs having Ruby dynamically resize the internal buffer of a single string (which iirc is 2x every time it needs to?).

@tenderlove do you have thoughts here?

ext/rugged/rugged_patch.c
tenderlove: Could we expose `git_patch_print` to Ruby?  Then we could implement this (and the other place that uses `git_patch_print`) in Ruby.

tenderlove:
> I do wonder about the trade-off there with generating more "garbage" (all the small strings). Being as though each of those strings in the array will be a new allocation vs having Ruby dynamically resize the internal buffer of a single string (which iirc is 2x every time it needs to?).

I guess it depends on the size of the diffs.  This iterator, is it per line?  Or something else?  If we expose the iterator to Ruby, we could open up other options, like streaming the diff.  We may not even have to worry about buffer performance at that point.

brianmario:
> This iterator, is it per line?

Yeah.

> If we expose the iterator to Ruby, we could open up other options, like streaming the diff.

Wouldn't that be slower because of all the C<->Ruby boundary crossing? Also, we don't support streaming where we'll be using this :\

tenderlove:
> Wouldn't that be slower because of all the C<->Ruby boundary crossing?

TBH, I'm not sure.  We could do both; expose the iterator and add use the C implementation in this PR.  Then we could test the perf difference.  Regardless of this PR, I suspect an iterator exposed to Ruby would be pretty handy.

> Also, we don't support streaming where we'll be using this :\

*sad trombone*

arthurschreiber:
> Regardless of this PR, I suspect an iterator exposed to Ruby would be pretty handy.

I'll add the iterator in a separate PR. For our current use case, I don't think we'll beat the C implementation for `to_s` or `header` in regards to performance - but it'll be really handy for any streaming related stuff. 👍

PR #630, opened 2016-09-07
Index.merge_file: raise when conflict is missing one side

ethomson:
libgit2's `git_merge_file` expects and `ours` and a `theirs` side so that it can either automerge the two, or if that fails, produce a merge file with conflicts marked up.  It cannot produce anything if the ours or the theirs side is missing.  (libgit2 in fact `assert`s here.)

Raise an error in this situation instead of trying to call `libgit2` where it will undoubtedly crash.

### DIFF
--- a/ext/rugged/rugged_index.c
+++ b/ext/rugged/rugged_index.c
@@ -1092,6 +1092,11 @@ static VALUE rb_git_merge_file(int argc, VALUE *argv, VALUE self)
 	else
 		rugged_exception_check(error);
 
+	if (ours == NULL)
+		rb_raise(rb_eRuntimeError, "The conflict does not have a stage 2 entry");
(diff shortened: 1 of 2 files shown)


ethomson:
Note that what Git does it to leave the one side that does exist in the working directory:

```
% git merge branch
CONFLICT (modify/delete): file.txt deleted in branch and modified in HEAD. Version HEAD of file.txt left in tree.
Automatic merge failed; fix conflicts and then commit the result.
```

So there's an argument to be made that rugged should return the one side that does exist.

I don't think that this is obvious, though, and I think that applications should instead decide what they want to do when only one of ours or theirs exist, and do so explicitly.  To call `Index.merge_files` is an error, IMHO.

vmg:
Seems reasonable!

PR #631, opened 2016-09-09
Patch to s timeout

mclark:
When calculating very large patches, there is the potential for the call to hang for a while without any feedback to the caller.

Potentially this could manifest itself in one of two places:

1. Calculating the diff_hunks in the original diff
2. Translating the lines of those hunks into ruby strings

This patch addresses the latter and gives us a mechanism to bail early when the diff output is taking too long.

I'm not sure if there is a clear path to addressing a timeout in the former case, but I am open to suggestions and feedback on both ideas.

Thanks so much to @brianmario for 🍐 ing with me on this one!

### DIFF
--- a/ext/rugged/rugged_patch.c
+++ b/ext/rugged/rugged_patch.c
@@ -23,9 +23,11 @@
  */
 
 #include "rugged.h"
+#include <sys/time.h>
 
 extern VALUE rb_mRugged;
 extern VALUE rb_cRuggedDiffDelta;
+extern VALUE rb_eRuggedTimeoutError;
 VALUE rb_cRuggedPatch;
 
 /*
@@ -327,13 +329,20 @@ static VALUE rb_git_diff_patch_bytesize(int argc, VALUE *argv, VALUE self)
 	return INT2FIX(bytesize);
 }
 
+struct patch_print_payload {
+	VALUE rb_buffer;
+	double timeout;
+	struct timeval start;
+};
+
 static int patch_print_cb(
 	const git_diff_delta *delta,
 	const git_diff_hunk *hunk,
 	const git_diff_line *line,
 	void *payload)
 {
-	VALUE rb_buffer = (VALUE)payload;
+	struct patch_print_payload *print_payload = (struct patch_print_payload*)payload;
+	VALUE rb_buffer = print_payload->rb_buffer;
 
 	switch (line->origin) {
 		case GIT_DIFF_LINE_CONTEXT:
@@ -343,6 +352,18 @@ static int patch_print_cb(
 	}
 
 	rb_ary_push(rb_buffer, rb_str_new(line->content, line->content_len));
+
+	if (print_payload->timeout > 0) {
+		struct timeval finish;
+		gettimeofday(&finish, NULL);
+
+		double elapsed_time = finish.tv_sec - print_payload->start.tv_sec;
+		elapsed_time += (double) (finish.tv_usec - print_payload->start.tv_usec) / 1000000.0;
+
+		if (elapsed_time >= print_payload->timeout) {
+			return GIT_EUSER;
+		}
+	}
 
 	return GIT_OK;
 }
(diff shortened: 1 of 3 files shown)


ext/rugged/rugged_patch.c
```diff
@@ -344,6 +353,18 @@ static int patch_print_cb(
 
 	rb_ary_push(rb_buffer, rb_str_new(line->content, line->content_len));
 
+	if (print_payload->timeout > 0) {
+		struct timeval finish;
+		gettimeofday(&finish, NULL);
```
vmg: Let's please use `clock_gettime` with `CLOCK_MONOTONIC` here.

ext/rugged/rugged_patch.c
```diff
@@ -365,19 +386,46 @@ static int patch_print_header_cb(
...
-	return rb_ary_join(rb_buffer, Qnil);
+	if (!NIL_P(rb_options)) {
+		VALUE rb_timeout = rb_hash_aref(rb_options, CSTR2SYM("timeout"));
+		if (!NIL_P(rb_timeout)) {
+			timeout = NUM2DBL(rb_timeout);
+		}
+	}
+
+	struct timeval start;
+	gettimeofday(&start, NULL);
```
vmg: Likewise.

vmg: Also, no mixed code/declarations, we're still C89 here.

vmg:
Looks pretty good. Let's switch to a monotonic clock please. :)

ext/rugged/rugged_patch.c
brianmario: @vmg we can probably just `#ifdef` this but `clock_gettime` doesn't exist on OSX until Sierra.

For now we could either stick with `gettimeofday` on OSX or `mach_absolute_time`? It may not work exactly the same as `clock_gettime`, but I think it's monotonic?

ethomson: FWIW, any [time functions in libgit2](https://github.com/libgit2/libgit2/blob/20302aa43738a972e0bd2e2ee6ae479208427b31/src/util.h#L598) use `clock_gettime(CLOCK_MONOTONIC...` when it's available and fall back to `gettimeofday`.  We never use `mach_absolute_time`, so that might be a precedent for you.

brianmario: @ethomson oh cool ok, let's just run with that then :)

@mclark we can pair on this again tomorrow if you'd like?

brianmario: @ethomson hah I actually just noticed https://github.com/libgit2/libgit2/blob/20302aa43738a972e0bd2e2ee6ae479208427b31/src/util.h#L565-L581

ext/rugged/time.h
```diff
@@ -0,0 +1,106 @@
...
+   if (scaling_factor == 0) {
+       mach_timebase_info_data_t info;
+       (void)mach_timebase_info(&info);
+       scaling_factor = (double)info.numer / (double)info.denom;
+   }
+
+   return (double)time * scaling_factor / 1.0E9;
+}
+
+#elif defined(AMIGA)
```
vmg: Rugged does not support AMIGA. :)

ext/rugged/time.h
```diff
@@ -0,0 +1,106 @@
...
+
+#include <sys/time.h>
+
+GIT_INLINE(double) rugged__timer(void)
+{
+	struct timespec tp;
+
+	if (clock_gettime(CLOCK_MONOTONIC, &tp) == 0) {
+		return (double) tp.tv_sec + (double) tp.tv_nsec / 1.0E9;
+	} else {
```
vmg: This branch can go. `clock_gettime` cannot fail on `CLOCK_MONOTONIC` as the source is required by POSIX.

ext/rugged/time.h
```diff
@@ -0,0 +1,106 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+ #ifndef __H_RUGGED_TIME__
+ #define __H_RUGGED_TIME__
+
+#if defined(_MSC_VER)
```
vmg: These macros are unnecessary. Let's stick to `static inline` without a macro.

ext/rugged/time.h
```diff
@@ -0,0 +1,106 @@
...
+# define GIT_INLINE(type) static __inline type
+#else
+# define GIT_INLINE(type) static inline type
+#endif
+
+#ifdef GIT_WIN32
```
vmg: This macro will never be defined on Rugged as it comes from libgit2's Makefile config.

The good news is that the whole branch can go, as it's a MSVC-specific branch that should not be necessary for Rugged, as it does not compile under MSVC.

vmg:
Looking good @mclark. As you can see, we can drop a lot of libgit2's original code because we support fewer platforms. That's good news!

Please do that and ping me back so I can audit the remaining code against libgit2's original blame -- as it is an OSS project, we cannot magically relicense it from GPLv2 to MIT without asking the contributors. Unless the contributors are @carlosmn and @ethomson. In that case we can, because we pay them money to abuse their copyright. 👌

mclark:
thanks @brianmario for catching that one! C is weird! 😛 

@vmg All concerns save the legal ones have been addressed. I can rebase this to clean it up if you like before we proceed.

mclark:
This PR is 7 years old, time to close!

PR #633, opened 2016-09-23
Introduce `Rugged::Blob.merge_files`

ethomson:
We have `Rugged::Index.merge_files` that will produce the results of a three-way merge for a conflict.  (It looks up the conflict in the index and passes the three sides to `xdiff`).

Provide a similar function that will take the file contents directly (either as a string, or with an oid) and perform the three-way merge and return the results.

Note that this function will accept either input as a string _or_ the given object ID of a blob to lookup and merge.  This may be a bit overwrought, TBH, as most callers probably only want to deal with oids...?

/cc @brianmario

### DIFF
--- a/ext/rugged/rugged_blob.c
+++ b/ext/rugged/rugged_blob.c
@@ -551,55 +551,105 @@ static VALUE rb_git_blob_to_buffer(int argc, VALUE *argv, VALUE self)
 	return rb_ret;
 }
 
-static void rugged_parse_merge_file_input(git_merge_file_input *input, VALUE rb_input)
+#define RUGGED_MERGE_FILE_INPUT_INIT { GIT_MERGE_FILE_INPUT_INIT }
+
+typedef struct {
+	git_merge_file_input parent;
+	int has_id;
+	git_oid id;
+} rugged_merge_file_input;
+
+static void rugged_parse_merge_file_input(rugged_merge_file_input *input, git_repository *repo, VALUE rb_input)
 {
 	VALUE rb_value;
 
 	Check_Type(rb_input, T_HASH);
 
-	rb_value = rb_hash_aref(rb_input, CSTR2SYM("content"));
-	if (NIL_P(rb_value))
-		rb_raise(rb_eArgError, "File input must have `:content`.");
+	if (!NIL_P(rb_value = rb_hash_aref(rb_input, CSTR2SYM("content")))) {
+		input->parent.ptr = RSTRING_PTR(rb_value);
+		input->parent.size = RSTRING_LEN(rb_value);
+	} else if (!NIL_P(rb_value = rb_hash_aref(rb_input, CSTR2SYM("oid")))) {
+		if (!repo)
+			rb_raise(rb_eArgError, "Rugged repository is required when file input is `:oid`.");
 
-	input->ptr = RSTRING_PTR(rb_value);
-	input->size = RSTRING_LEN(rb_value);
+		rugged_exception_check(git_oid_fromstr(&input->id, RSTRING_PTR(rb_value)));
+		input->has_id = 1;
+	} else {
+		rb_raise(rb_eArgError, "File input must have `:content` or `:oid`.");
+	}
 
 	rb_value = rb_hash_aref(rb_input, CSTR2SYM("filemode"));
 	if (!NIL_P(rb_value))
-		input->mode = FIX2UINT(rb_value);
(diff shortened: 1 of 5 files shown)


ethomson:
@arthurschreiber Objections?

ext/rugged/rugged_index.c
```diff
@@ -1100,11 +1027,7 @@ static VALUE rb_git_merge_file(int argc, VALUE *argv, VALUE self)
 	error = git_merge_file_from_index(&merge_file_result, repo, ancestor, ours, theirs, &opts);
 	rugged_exception_check(error);
 
-	rb_hash_aset(rb_result, CSTR2SYM("automergeable"), merge_file_result.automergeable ? Qtrue : Qfalse);
-	rb_hash_aset(rb_result, CSTR2SYM("path"),         rb_path);
-	rb_hash_aset(rb_result, CSTR2SYM("filemode"),     INT2FIX(merge_file_result.mode));
-	rb_hash_aset(rb_result, CSTR2SYM("data"),         rb_str_new(merge_file_result.ptr, merge_file_result.len));
-
+    rb_result = rb_merge_file_result_fromC(&merge_file_result);
```
arthurschreiber: There are spaces here instead of a tab.

ext/rugged/rugged_blob.c
```diff
@@ -551,6 +551,88 @@ static VALUE rb_git_blob_to_buffer(int argc, VALUE *argv, VALUE self)
...
+		Data_Get_Struct(rb_repo, git_repository, repo);
+	}
+
+	if ((error = rugged_parse_merge_file_input(&ancestor, &ancestor_blob, repo, rb_ancestor)) < 0 ||
+		(error = rugged_parse_merge_file_input(&ours, &our_blob, repo, rb_ours)) < 0 ||
+		(error = rugged_parse_merge_file_input(&theirs, &their_blob, repo, rb_theirs)) < 0)
+		goto done;
+
+	if (!NIL_P(rb_options))
+		rugged_parse_merge_file_options(&opts, rb_options);
```
arthurschreiber: You'll want to call `rugged_parse_merge_file_options(&opts, rb_options)` before doing any of the `rugged_parse_merge_file_input` calls.

If a ruby exception gets raised inside `rugged_parse_merge_file_options` (e.g. by `Check_Type`), `ancestor`, `ours` and `theirs` will not be freed again.

ext/rugged/rugged_blob.c
```diff
@@ -551,6 +551,88 @@ static VALUE rb_git_blob_to_buffer(int argc, VALUE *argv, VALUE self)
...
+	int error;
+
+	rb_scan_args(argc, argv, "41", &rb_repo, &rb_ancestor, &rb_ours, &rb_theirs, &rb_options);
+
+	if (!NIL_P(rb_repo)) {
+		rugged_check_repo(rb_repo);
+		Data_Get_Struct(rb_repo, git_repository, repo);
+	}
+
+	if ((error = rugged_parse_merge_file_input(&ancestor, &ancestor_blob, repo, rb_ancestor)) < 0 ||
```
arthurschreiber: I see that `rugged_parse_merge_file_input` might also raise ruby exceptions. This can also lead to memory leaks, as noted below.

One way to work around that would be to break this function up into two parts, one that does input validation (and may raise ruby exceptions) and the other that performs the actual actions (and does not raise exceptions).

brianmario:
> This may be a bit overwrought, TBH, as most callers probably only want to deal with oids...?

Yeah I keep going back and forth on this myself. I can definitely think of valid use cases for accepting both types of parameters but it is a pain to continue to do that everywhere and muddies up the API a bit.

If we were to go to an almost pure object-parameter-only API in the future I think we'd just want to make sure we're avoiding allocations wherever possible. Because it could just be that the caller only wants an OID string out of it.

scottjg:
@ethomson hey do you have a rough idea when this might land? it's blocking the conflict resolver now - i may just take the in progress work from this PR and deploy it if you think it might be more than a few days?

arthurschreiber:
I can wrap this up and get it landed, maybe today.

ethomson:
Hey @scottjg - sorry, I was out of the office.  I'm back and will fix this up directly!

ethomson:
Thanks for the review @arthurschreiber - I fixed this up and indeed pulled it into two separate functions, one that parses the options and a second that actually loads the blobs.  Only the first should be able to `raise`, while the second explicitly returns libgit2 error codes.

Issue #636, opened 2016-10-05
rb_git_indexentry_toC randomly setting `assume-unchanged` flag?


jasonroelofs:
I have a really strange situation that may be a bug or just confusion on my part. Basically, I use Rugged to create files in a repository using the following code:

```ruby
oid = @repo.write(content, :blob)
index.add(path: path, oid: oid, mode: 0100644)
tree_oid = index.write_tree(@repo)
# Create Rugged::Commit with the tree_oid and write out to disk
```

For the most part this has been working fine but I've realized that when I go into the repository itself (the repo is not bare) and start using `git` commands, some files are flagged as `assume-unchanged`, such that local changes do not show up in `git status`. E.g., when I run `git ls-files -v` on files in the repo I see the following:

```
...
H themes/high-fiber/raw_page.tpl
h themes/high-fiber/sitemap.xml.tpl
h themes/high-fiber/stylesheets/base.css
...
```

Where `git ls-files -v` does the following:

```
-v
    Similar to -t, but use lowercase letters for files that are marked as assume unchanged (see git-update-index(1)).
```

I was led to this information when I found this SO answer: http://stackoverflow.com/a/28657085 and so I started diving down into how Rugged and libgit2 were writing these files, because there feels like a disconnect between how git reads index flags and how rugged is setting said flags.

My investigation led me to `rb_git_indexentry_toC` in `rugged_index.c`, specifically near the end of that function where `flags` is set:

```c
	if (!NIL_P(val)) {
		entry->flags &= ~GIT_IDXENTRY_VALID;
		if (rugged_parse_bool(val))
			entry->flags |= GIT_IDXENTRY_VALID;
	} else {
		entry->flags |= GIT_IDXENTRY_VALID;
       }
```

Now, in git, the flag for "assume-unchanged" is `0x8000` and called `CE_VALID`, which was ported to libgit2 as this `GIT_IDXENTRY_VALID`. However, all of the code in git that I've been able to find says that if this flag is set, then that means "assume-unchanged", so "valid" means "I will never change". I'm wondering if that expectation was confused here, and should the 'else' case in the code above be removed (as flags is initialized to 0)?

This `else` clause was added in https://github.com/libgit2/rugged/commit/771b784f44ce20dbbed7eaca0e659743d3add253.

[...]

PR #638, opened 2016-10-06
Don't default index entries to "VALID"

jasonroelofs:
The "VALID" flag in git refers to the "assume-unchanged" feature, such
that if, in a non-bare repository, an index entry is marked as "VALID",
the git CLI (e.g. `git status`) will assume that the file never changes,
and ignore any changes that do occur.

See https://github.com/libgit2/rugged/issues/636 for the issue and research leading to this change.

I wasn't able to track down a good way of unit testing this change, as
these flags aren't really exposed in Rugged, and I didn't notice any tests
working on non-bare repositories.


logicminds:
Can we get this merged I am running into this issue as well.  Thanks for documenting @jasonroelofs  so I can have a work around.

Nothing is random for me.  It happens every time.

ethomson:
Not only is this correct, we should change the option name from `valid` to `assume_unchanged` to clarify its actual behavior.

PR #571, opened 2016-02-10
Added support for certificate_check callback

mkanoor:
Since libgit2 removed support for GIT_SSL_NO_VERIFY environment variable
and  "http.sslverify" in favor of certificate check callbacks, we need
rugged to support the certificate check callback.

One usecase for this is self-signed certificates which can be verified
in the callback.

PR #639, opened 2016-10-06
Remote: Expose `certificate_check` callback

arthurschreiber:
This exposes the `certificate_check` callback for `Rugged::Remote#fetch`.

It is based on #571, but changes the `valid` parameter to be `true` or `false` instead of `1` or `0`, and cleans up some of the code and the test cases.

Fixes #571.


kylemacey:
I think the docs need to be updated for https://www.rubydoc.info/github/libgit2/rugged/Rugged%2FRepository.clone_at

PR #642, opened 2016-10-14
Move Rugged::Index#diff to Ruby

tenderlove:
This PR moves most of the branching logic in `Rugged::Index#diff` to Ruby.  Since each private C function only exposes one libgit2 function, it will be easier to unlock the GVL around that one function.


ext/rugged/rugged_index.c
```diff
@@ -691,146 +691,51 @@ static VALUE rb_git_index_readtree(VALUE self, VALUE rb_tree)
...
-		} else {
-			xfree(opts.pathspec.strings);
-			rb_raise(rb_eTypeError, "A Rugged::Commit or Rugged::Tree instance is required");
-		}
-	}
+	// Need to flip the reverse option, so that the index is by default
+	// the "old file" side of the diff.
+	opts.flags ^= GIT_DIFF_REVERSE;
+
+	git_tree *other_tree;
```
arthurschreiber: Can you move this declaration to the top of the function?

arthurschreiber:
The test failures are happening due to building with the wrong libgit2 version. I'm not really sure what is going on, but as long as the travis build is green, this should be good to be merged.

ext/rugged/rugged_index.c
tenderlove: oops!  Yes, done.

PR #644, opened 2016-10-25
Add regression tests for error cases on `Rugged::Diff.diff`

tenderlove:
I would like to start refactoring `Rugged::Diff.diff`, but we are
missing tests for some of the branches in the C code.  This commit adds
tests for those branches.

I wasn't sure where to put these tests, so I just made a new test class.


vmg:
👌

PR #647, opened 2016-10-26
Fix memory leak in `Rugged::Tree.diff`

tenderlove:
If `Rugged::Tree.diff` can't unwrap the repository object passed in, it
can leak memory allocated in `rugged_parse_diff_options`.  Before this
commit, memory usage for this script will grow unbounded:

```ruby
require 'rugged'

loop do
  begin
    Rugged::Tree.diff nil, nil, nil, { paths: ['x' * 90] }
  rescue TypeError
  end
end
```

To fix this, just try to unwrap the repository object before calling
`rugged_parse_diff_options`.


vmg:
:pray:

PR #649, opened 2016-10-30
Fix misprint and incorrect formatting + remove obsolete license headers

apoluektov:


arthurschreiber:
@vmg Do we want to remove the copyright notices from the individual c files?

ethomson:
There probably shouldn't be _nothing_.

I concede that changing the date is quite obnoxious.  What libgit2 did here (a brief, dateless copyright blurb) seems like the least common denominator that lawyers and devs seem happy with.

apoluektov:
But what is the difference between rugged *.rb and *.c files regarding to this issuse? Why the former don't require the copyright notices but the latter do?

I though as long as the LICENSE file is here, and its content is identical to the copyright notice, there were no need to repeat it in the individual files... Am I wrong?

ethomson:
IANAL, but there's no explicit _requirement_ per se.  Speaking generally, a few lines at the top of one source file isn't necessary, rugged is copyrighted regardless.

My suggestion would be:
1. Create a brief statement.
2. Omit date and specific copyright owners from this statement, since those are the things that might need changing.  (And either require maintenance or suggest obsolescence.)
3. Apply it to all the source files, be they C or Ruby.

Modifying the libgit2 blurb would give us:
```
/*
 * Copyright (C) the Rugged contributors.  All rights reserved.
 *
 * This file is part of Rugged, distributed under the MIT license.
 * For full terms see the included LICENSE file.
 */
```

apoluektov:
Ok, sounds great!

If there are no other objections, I'll proceed with that.

apoluektov:
Done. I hope we don't need to add those headers to tests/scripts :)

apoluektov:
bump

vmg:
Looking good! Thanks for the cleanup :)

PR #650, opened 2016-11-02
Unlock GVL around `git_diff_tree_to_tree`

tenderlove:
This lets us execute Ruby code in parallel while making diffs.

### DIFF
--- a/ext/rugged/rugged_tree.c
+++ b/ext/rugged/rugged_tree.c
@@ -23,6 +23,7 @@
  */
 
 #include "rugged.h"
+#include <ruby/thread.h>
 
 extern VALUE rb_mRugged;
 extern VALUE rb_cRuggedObject;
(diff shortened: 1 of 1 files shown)


arthurschreiber:
@tenderlove What's the long term plan here? Do we want to wrap all the "expensive" libgit2 calls in GVl-unlock code?

carlosmn:
We're going to try this out and see how much of an improvement we can get by being able to do things like timeouts without having the thread that's counting time have to wait for the libgit2 call to be over. Many calls can probably be run outside the GVL but we're going to take them piecemeal.

tenderlove:
@arthurschreiber that's what I would like to do.  Though I don't know which APIs are "slow".  I only wrapped this function because I was able to demonstrate that it's slow.  I'm not sure what other APIs need wrapping, but once we do a second or third one, I think we can eliminate some of this boilerplate C code with some macros.

PR #679, opened 2017-02-08
Allow disabling rename detection for merge operations.

arthurschreiber:
The default for rename detection for merge operations changed in the latest libgit2 version from being disabled to being enabled. This change allows users to specify `renames: false` to disable this detection.

### DIFF
--- a/ext/rugged/rugged_tree.c
+++ b/ext/rugged/rugged_tree.c
@@ -487,8 +487,8 @@ void rugged_parse_merge_options(git_merge_options *opts, VALUE rb_options)
 			}
 		}
 
-		if (RTEST(rb_hash_aref(rb_options, CSTR2SYM("renames")))) {
-			opts->flags |= GIT_MERGE_FIND_RENAMES;
+		if (rb_hash_aref(rb_options, CSTR2SYM("renames")) == Qfalse) {
+			opts->flags &= ~GIT_MERGE_FIND_RENAMES;
 		}
 
 		if (RTEST(rb_hash_aref(rb_options, CSTR2SYM("fail_on_conflict")))) {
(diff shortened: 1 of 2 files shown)


ext/rugged/rugged_tree.c
```diff
@@ -489,6 +489,8 @@ void rugged_parse_merge_options(git_merge_options *opts, VALUE rb_options)
 
 		if (RTEST(rb_hash_aref(rb_options, CSTR2SYM("renames")))) {
 			opts->flags |= GIT_MERGE_FIND_RENAMES;
+		} else {
+			opts->flags &= ~GIT_MERGE_FIND_RENAMES;
```
ethomson: You could drop the first test, since this is set already, and just `if (!RTEST...)` and unset that bit.

ethomson: Entirely optional, of course.  I'm going to 👍 this either way of course.  😄

arthurschreiber: `!RTEST` won't work here because `nil` also evaluates to `0` with `RTEST` and not specifying this option would clear it 😖.

ext/rugged/rugged_tree.c
```diff
@@ -487,8 +487,8 @@ void rugged_parse_merge_options(git_merge_options *opts, VALUE rb_options)
 			}
 		}
 
-		if (RTEST(rb_hash_aref(rb_options, CSTR2SYM("renames")))) {
-			opts->flags |= GIT_MERGE_FIND_RENAMES;
+		if (rb_hash_aref(rb_options, CSTR2SYM("renames")) == Qfalse) {
```
ethomson: Ah of course.  Sorry, I now realize that my prior suggestion doesn't make sense.  😀

PR #680, opened 2017-02-08
docs: Iterator -> Enumerator

alyssais:
There's no such thing as an `Iterator` in Ruby. What these methods actually return is an `Enumerator`.


alyssais:
That weird open/close thing looks like it happened because of [this](https://twitter.com/githubstatus/status/829379605702254592).

alyssais:
CI failure looks related to that too.

arthurschreiber:
Thanks! ❤️

PR #687, opened 2017-03-03
Update libgit2 to b31b236...

ethomson:
/cc @arthurschreiber @carlosmn


ethomson:
It appears that we broke reference namespaces a little bit in libgit2?  I'll dig in this weekend.

ethomson:
I fixed the namespace issues in libgit2 and updated this PR to include those changes.

ethomson:
@carlosmn One more set of 👀 if you don't mind?  I turned on the SHA-1 collision detection by default (and rebased this on master).

PR #686, opened 2017-03-01
On the way to fix / improve the Windows build

sschuberth:
@arthurschreiber @carlosmn Please have a look at the individual commit messages for the details.

PR #689, opened 2017-03-06
Move determining the make executable to earlier in the build process

sschuberth:
This way we can force rake-compiler to use whatever we found instead of
letting it determine the make executable via its own rules again.


sschuberth:
@arthurschreiber This is a follow-up to PR #686.

sschuberth:
Closing this due to a lack of interest by the maintainers.

PR #690, opened 2017-03-07
add a little hint in the readme for google

phoet:
i think it would be helpful to find the installation instructions of the readme in case of the `ERROR: CMake is required to build Rugged.`. 

i just bundle installed a new repo and it failed with this error, so it's not apparent that one has to look into how to install rugged properly. i eventually found [this stackoverflow answer](http://stackoverflow.com/a/36587267/100731) but not the readme itself via google.


vmg:
Thank you :)

PR #704, opened 2017-07-07
Update rubinus name for Travis

carlosmn:


carlosmn:
We gave up on testing against rubinius some time ago.

Issue #701, opened 2017-07-06
Repository#status is hard to use.


ioquatix:
Here is an example:

```ruby
changes = {}
repository.status do |file, status|
	unless status == [:ignored]
		changes[file] = status
	end
end

next if changes.empty?

logger.info "Package #{package.name} (from #{package.path}):".bright

changes.each do |file, status|
	if status == [:worktree_new]
		logger.info "\t#{file}".color(:blue)
	elsif status == [:worktree_modified]
		logger.info "\t#{file}".color(:orange)
	elsif status == [:worktree_deleted]
		logger.info "\t#{file}".color(:red)
	else
		logger.info "\t#{file} #{status.inspect}"
	end
end
```

I firstly have to assemble the list of changes into a hash, then check if it's empty, before finally printing out details.

It would be nice if you could write `changes = repository.status` or even better, if repository.status could return a chainable enumerable `to_enum`.

ioquatix:
So, I could write, for example:

```ruby
changes = repository.status.select{|file, status| status != [:ignored]}
```

ioquatix:
Maybe even better would be if the API could filter out this data before even coming to Ruby-land:

```ruby
changes = repository.status(ignored: false)
```

ioquatix:
Here is another example of what I think would be nice to do:

```ruby
# Check dirty status of local repository:
if context.repository.status(ignored: false).any?
    abort "You have unstaged changes/unadded files. Please stash/commit them before running the generator.".color(:red)
end
```

PR #714, opened 2017-08-01
Move Rugged::Repository#status to Ruby

tenderlove:
I think this will make other changes like in #701 much easier, and
removes some branching code from C.


carlosmn:
Makes sense to me.

PR #715, opened 2017-08-01
Fix memory leak in Rugged::Repository#status

tenderlove:
`Rugged::Repository#status` takes a block, and the block form of the
method will call `git_status_foreach`.  That function will eventually
[allocate memory](https://github.com/libgit2/libgit2/blob/a9d6b9d529daa920b558ee3f34151422d10bc95e/src/status.c#L425),
and it doesn't free the memory until after the `git_vector_foreach`
function finishes iteration.  The problem is that anything that does a
long jump (like raising an exception, or calling `break`) will jump out
of the `git_status_foreach_ext` before the list gets freed, so we will
have a leak.  For example, this program will consume memory
indefinitely:

```ruby
require 'rugged'

repo = Rugged::Repository.new ARGV[0]
loop do
  repo.status do |file, status_data|
    break
  end
end
```

This commit ensures the loop finishes and memory gets freed before any
long jumps happen.  Note that this is a **HUGE HACK**.  To make sure the
code jumps to the right place, I took advantage of the fact that Rugged
error codes are less than 0 and Ruby jump tags are greater than 0, and
that the `git_vector_foreach` will return the error code if it is
anything *except* 0.  This means I can overload the error code to be the
Ruby jump destination.  I apologize to the programming gods for my
transgressions, but this seemed like the easiest way to return the jump
location back up the stack.

### DIFF
--- a/ext/rugged/rugged_repo.c
+++ b/ext/rugged/rugged_repo.c
@@ -1504,11 +1504,12 @@ static VALUE flags_to_rb(unsigned int flags)
 
 static int rugged__status_cb(const char *path, unsigned int flags, void *payload)
 {
-	rb_funcall((VALUE)payload, rb_intern("call"), 2,
-		rb_str_new_utf8(path), flags_to_rb(flags)
-	);
+	int exception = GIT_OK;
 
-	return GIT_OK;
+	rb_protect(rb_yield, rb_ary_new3(2, rb_str_new_utf8(path),
+		    flags_to_rb(flags)), &exception);
+
+	return exception;
 }
 
 /*
@@ -1575,7 +1576,12 @@ static VALUE rb_git_repo_status(int argc, VALUE *argv, VALUE self)
 		(void *)rb_block_proc()
 	);
 
-	rugged_exception_check(error);
+	if (error > 0) {
+	    rb_jump_tag(error);
+	} else {
+	    rugged_exception_check(error);
+	}
+
 	return Qnil;
 }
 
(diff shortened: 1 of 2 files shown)


arthurschreiber:
Thanks for the fix! ❤️ I'll take a closer look in the next few days and see what other parts of rugged do to not leak memory in similar cases.

ext/rugged/rugged_repo.c
```diff
@@ -1504,11 +1504,12 @@ static VALUE flags_to_rb(unsigned int flags)
 
 static int rugged__status_cb(const char *path, unsigned int flags, void *payload)
 {
-	rb_funcall((VALUE)payload, rb_intern("call"), 2,
-		rb_str_new_utf8(path), flags_to_rb(flags)
-	);
+	int exception = GIT_OK;
```
carlosmn: This is confusingly named, since it mostly should not be, if would be less confusing if this were a literal 0 since that's what `exception` will be if everything goes well. We can then check whether `exception` is non-0 and then return whatever we feel is appropriate. Let's not be stingy with the number of variables.

carlosmn: Looking at this more in context I see now why it's so awkward as this is expecting a libgit2 return code but we need to preserve the exception id or whatever `rb_protect` returns.

On the one hand we should switch to a better iterator via `git_status_list` but keeping with this kind of loop, this is what `GIT_EUSER` is for. We guarantee that we pass this along and that the library won't generate this. You can then use the payload (or whatever other mechanism you have) to store whatever information and if you see `GIT_EUSER` returned, then you _know_ that this was generated by you and you can look up the information in the payload that you passed to the foreach function.

carlosmn:
I've opened an alternative with #787 which, like, actually makes use of the library the way it expects you to.

This is a neat find, I suspect some of the unexplained memory consumption issues come from this kind of call.

carlosmn:
I've merged #787 instead. Thanks for the heads up on the issue.

PR #730, opened 2018-01-03
Fix for returning enumerators.

### LINKED ISSUE #729: `each_line` causes segfault with ruby 2.5.0 MRI
I was investigating why [pronto](https://github.com/prontolabs/pronto) fails with some weird errors around [this line](https://github.com/prontolabs/pronto/blob/6d21d1f4991f4f3f3871217bd034536cae93d915/lib/pronto/git/patch.rb#L22) and after some debugging came to the conclusion that there is an issue between the current version of rugged and ruby 2.5.0.

It's easy enough to reproduce, just clone the repo as described in the docs and try running the test suite. It'll fail, causing a segfault. During the compilation process, several "division by zero" warnings are thrown.

My environment:

* macOS 10.13.1
* cmake 3.10.1
* ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin17]

I'll try running the test suite under ubuntu as well and will update this issue accordingly.

Logs/traces:

[compile.log](https://github.com/libgit2/rugged/files/1597717/compile.log)
[test_run.log](https://github.com/libgit2/rugged/files/1597718/test_run.log)

arthurschreiber:
In https://github.com/libgit2/rugged/issues/729, a user described issues with using rugged together with the recently released Ruby 2.5.0.

The cause of the problems was that we sometimes passed more arguments to `rb_funcall` than we specified, causing segfaults. `rb_funcall` was used to return enumerator objects when iterator methods got called without a block being passed in.

While looking into fixing this, I stumbled upon the `RETURN_ENUMERATOR` and `RETURN_SIZED_ENUMERATOR` macros provided by Ruby, which seem like a more elegant solution to return enumerator objects.

This fixes #729.


arthurschreiber:
Looks like Ruby 2.5.0 is currently somewhat broken in Travis CI, so I reverted 6b335ebbc8e975e51d450701d1d3a819e1d01ed8. We can add Ruby 2.5.0 builds in a later PR once the issue has been resolved. 👍

npetkov:
Take a look at [this travis issue](https://github.com/travis-ci/travis-ci/issues/8978), just completed a successful build with `gem update --system --no-doc` and can confirm it works.

carlosmn:
The last argument should be `NULL` rather than `0` as it's a pointer. Or is this the way the function is called generally?

arthurschreiber:
> The last argument should be NULL rather than 0 as it's a pointer. Or is this the way the function is called generally?

I can change this, as it shouldn't make a difference in the way the macro handles the argument but at the same time convey more context. But passing `0` for `argv` mirrors the way this macro is used in the ruby source. 🤷‍♂️

carlosmn:
Let's leave it then, we're unlikely to run on any machine where the zero pointer is actually a different value from the number zero.

Do you know what's up with the macOS 2.0.0 build? Can we expect it to recover?

npetkov:
I’m not sure why the 2.0.0 build fails, there seem to be [plenty of binaries](https://rvm.io/binaries/osx/10.12/x86_64/) for 10.12. Confusingly enough, the latest version is p647, not p648 which should be the last released 2.0 version.

On a side note, does it make sense to test against ruby 2.0 in the first place? It’s a rather old version and isn’t officially supported for almost two years now.

arthurschreiber:
@npetkov Yeah, the travis config is a bit out of date. I _think_ we can drop the builds on unsupported ruby versions.

carlosmn:
Anything older than 2.2 is EOL, so I wouldn't feel bad dropping support for those. Though let's do that in a different PR so we don't conflate too much.

carlosmn:
This should pass with 2.5.0 on Travis now, just as soon as some macOS machines are free.

npetkov:
@carlosmn Please see my first comment above.

carlosmn:
Let's cross-reference it better. #732 updates us to run against 2.5 as well as removes a few older versions and adds the workaround of updating the system rubygems.

PR #731, opened 2018-01-05
Add clone_at2 and fetch2

ghost:
Current version of clone_at and fetch dont release GVL when transfering data between client and server, so I just simply dup clone_at and fetch to remove their GVL and rename to clone_at2 and fetch2.

Limit:
The internal data structure in libgit2 is not protected by locks so dont use these APIs in mutlithread.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ addons:
     - openssh-client
     - openssh-server
 
-sudo: false
+sudo: required
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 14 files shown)


arthurschreiber:
Have you seen that there is a `rb_thread_call_with_gvl` which allows reaquiring the GVL to call ruby C-API functions again? I think using that you could keep the push notification callback behaviour as-is and not require `khash`.

scutuatua-crypto (approved):

scutuatua-crypto (approved):

PR #733, opened 2018-01-06
vendor: update libgit2 to a7e36d3d9c47

carlosmn:


carlosmn:
Linking issues on Ubuntu and NFD ref issues in macOS. cc libgit2/libgit2#4465

carlosmn:
On the macOS side, some CMake rule changes made us stop finding iconv when building a static archive. libgit2/libgit2#4467 tries to deal with the issue.

carlosmn:
The `.a` change made us work on macOS again, but on Ubuntu we still have the issue with not finding the SSL library. It did work when using the branch directly, so it's rather confusing.

PR #734, opened 2018-01-08
Wrap trailer parsing API

brianmario:
This wraps the new [trailer parsing API](https://github.com/libgit2/libgit2/pull/4451) up in a new method `Rugged::Commit#trailers`. The return value is an array of arrays like so:

```
[
  ["Trailername", "value"],
  ...
]
```

The idea here is to maintain the order the parser saw the trailers, and let the caller merge them into a hash if they don't care about the order.

We're still working on getting the libgit2 side merged for this, but this side shouldn't require any changes once that's ready.

cc @charliesome @arthurschreiber


vmg:
This is looking good. Definitely liking the array approach to ensure we keep the ordering on the headers.

PR #756, opened 2018-07-15
Set a custom allocator to redirect via the ruby runtime

carlosmn:
This lets us tell the ruby runtime how much memory rugged/libgit2 is allocating to give it a more realistic view of how much memory the process is using.
    
The expectation/hope is that long-lived processes will give back memory to the system more readily than is currently the case.

---

This is something which was spoken about at the contributors' summit as a likely source of server-side processes using rugged taking up silly amounts of memory.

/cc @arthurschreiber who suggested this

### DIFF
--- a/ext/rugged/rugged_allocator.c
+++ b/ext/rugged/rugged_allocator.c
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) the Rugged contributors.  All rights reserved.
+ *
+ * This file is part of Rugged, distributed under the MIT license.
+ * For full terms see the included LICENSE file.
+ */
+
+#include "rugged.h"
+#include <git2/sys/alloc.h>
+
+static void *rugged_gmalloc(size_t n, const char *file, int line)
+{
+	return xmalloc(n);
+}
+
+static void *rugged_gcalloc(size_t nelem, size_t elsize, const char *file, int line)
+{
+	return xcalloc(nelem, elsize);
+}
+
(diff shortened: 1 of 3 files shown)


ext/rugged/rugged_allocator.c
```diff
@@ -0,0 +1,97 @@
...
+{
+	char *newstr;
+	size_t n;
+
+	n = strlen(str);
+	newstr = xmalloc(n+1);
+	memcpy(newstr, str, n);
+	newstr[n] = '\0';
+
+	return newstr;
```
tenderlove: [Ruby implements a strdup](https://github.com/ruby/ruby/blob/e75fbfdacbf7d064dfcd908e222418a2da3869a5/util.c#L495-L505) so we can use that.

carlosmn: I saw that, but it wouldn't find it when building, so I thought it was internal. Do we need to add an `#include`?

tenderlove: Ya, it's public.  We just have to `#include <ruby/util.h>`

arthurschreiber: Done via https://github.com/libgit2/rugged/pull/756/commits/f049f6d5f3fd12dc0d3a138239b308f7fc694aed.

carlosmn:
Thanks for the update. I'm still some time off from being able to put this into production, but we can play around with a test environment.

PR #762, opened 2018-08-27
Use strnlen in the strndup allocator

carlosmn:
If we use `strlen` we'll go over the whole string rather than just the part that
interests us, and sometimes we are given large strings.

As en example, a 15MB reflog took ~5s to parse when using `strlen` and it fell
within the noise of the tests when using `strnlen`.


carlosmn:
The callsite should probably be using `git__substrdup` to be honest to avoid the `strlen` altogether, but this is still something we should make efficient.

PR #763, opened 2018-08-29
Give CMake a timeout for when it hangs

carlosmn:
In Debian Jessie, CMake will sometimes fail to correctly wait for the programs
it has spawned. While a proper fix would be in CMake itself, protect ourselves
from this by giving it a generous but limited time in which to run.

When it hangs, it will now instead be killed by the build process and we can try
again, instead of having to hunt down processes across machines.

---

I'm not sure how widespread this is, but on newer CMake versions the usual workaround for CMake hanging on Jessie is not working anymore. This limits the damage by actually timing out in error instead of just hanging forever.

I'm sending this as it probably doesn't hurt to have it upstream.

### DIFF
--- a/ext/rugged/extconf.rb
+++ b/ext/rugged/extconf.rb
@@ -4,6 +4,7 @@
 # For full terms see the included LICENSE file.
 
 require 'mkmf'
+require 'timeout'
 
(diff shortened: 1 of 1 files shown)


carlosmn:
Failures here also due to the message change fixed in #781

PR #769, opened 2018-11-09
Support Ruby's convention for pathnames and Pathname lib

marcandre:
Rugged is currently too strict in not recognizing objects that `respond_to? :to_path`. In particular, this means we can't pass a `Pathname` object directly, contrary to all builtin methods accepting paths for example.

This PR fixes the cases.

I simply searched for all `Check_Type(.*_path, T_STRING)`, so I may have missed some.


tenderlove:
This is güd.  Thanks @marcandre

PR #774, opened 2018-12-07
Allow diff_patch to run without the global lock

stanhu:
We noticed that our service would pause and refuse to service any other
requests when a large commit was requested as a patch. This allows
concurrent requests to be processed.


stanhu:
@carlosmn Would you mind reviewing this?

stanhu:
@carlosmn Thanks! Is there a timeline for a new release for Rugged?

carlosmn:
Rugged is released in lockstep with libgit2 when we update the vendored version of the library.

ext/rugged/rugged_diff.c
```diff
@@ -197,6 +198,20 @@ static int diff_print_cb(
...
+	git_diff_format_t format;
+	VALUE rb_str;
+};
+
+static void rb_git_diff_patch_nogvl(void * _args)
+{
+	struct nogvl_diff_patch_args * args;
+
+	args = (struct nogvl_diff_patch_args *)_args;
+	git_diff_print(args->diff, args->format, diff_print_cb, (void*) args->rb_str);
```
tenderlove: `diff_print_cb` calls `rb_str_cat`, and it's not safe to call `rb_str_cat` while the GVL is unlocked.  `rb_str_cat` can possibly resize the underlying string buffer, which can cause the GC to execute.  The GC is not thread safe, so this could cause havoc on the mutator.

carlosmn: https://github.com/ruby/ruby/blob/8b3774be3dd9f472bddd99e84e3c9fe2ff99d7ac/thread.c#L1548-L1551 says reallocations should be same as they acquire the GVL when needed.

carlosmn:
We've confirmed this change has lead to some lockups, likely related to what @tenderlove mentioned above about allocating memory, though we haven't pinpointed the specific cause.

As it causes the ruby vm to become unresponsive, we'll have to revert this.

pravi:
@carlosmn @stanhu  what is the current status of this? Is a different fix applied for this problem?

PR #781, opened 2019-02-15
test: update certificate rejection message

carlosmn:


carlosmn:
The errors are due to 2.2 not working on Travis which is fixed by removing it in #782

PR #782, opened 2019-02-15
travis: drop 2.2 and include 2.6

carlosmn:
The 2.2 version isn't working on Travis anymore and it's too old for us to care
about anyway.

Since the 2.6 version has been released, include that one while here.


carlosmn:
The failures are due to the error message that we're fixing with #781

PR #784, opened 2019-02-27
Bump libgit2 to v0.28.1

stanhu:


carlosmn:
Thanks, we had missed creating the branch for this version so I changed the target.

PR #788, opened 2019-03-12
Add Repository#apply

dometto:
Initial support for the `git_apply` functionality added to `libgit2` in https://github.com/libgit2/libgit2/pull/4705/. This is my first time working on `rugged`, so I'm happy to do whatever is needed to get this PR up to standards for the project.

Call signature: `Repository#apply(diff, options = {})`.
Options:
```
:location - Whether to apply the changes to the index, the working directory, or both. Valid values: :index, :workdir, :both. Defaults to :both.
```

At the moment, `#apply` returns `true` on success. On failure, an exception will be thrown (because of the call to `rugged_exception_check(error)`), which means in practice `false` will never be returned. I thought about returning the modified index instead, but that doesn't make sense given the possibility of modifying only the workdir.

Since `options` at the moment only allows for `:location`, another way to go would be to change the second argument to an optional `location` parameter. However, if other options are to be added in the future, it would make sense to set `:location` in the options Hash. (I am not sure whether it would make sense to add support for the `hunk_cb` and `delta_cb` options implemented [here](https://github.com/libgit2/libgit2/blob/4e746d80d229a77b08c5689a64d880bde5fd960f/include/git2/apply.h#L67-L68))

### DIFF
--- a/ext/rugged/rugged_repo.c
+++ b/ext/rugged/rugged_repo.c
@@ -970,8 +970,9 @@ static VALUE rb_git_repo_revert_commit(int argc, VALUE *argv, VALUE self)
  *  The following options can be passed in the +options+ Hash:
  *
  *  :location ::
- *    Whether to apply the changes to the workdir (default),
- *    the index, or both. Valid values: +:index+, +:workdir+, +:both+.
+ *    Whether to apply the changes to the workdir (default for non-bare),
+ *    the index (default for bare) or both. Valid values: +:index+, +:workdir+,
+ *    +:both+.
  *
  *  :delta_callback ::
  *    While applying the patch, this callback will be executed per delta (file).
--- a/ext/rugged/rugged_repo.c
+++ b/ext/rugged/rugged_repo.c
@@ -10,6 +10,7 @@
 #include <git2/sys/odb_backend.h>
 #include <git2/sys/refdb_backend.h>
 #include <git2/refs.h>
+#include <git2/apply.h>
 
 extern VALUE rb_mRugged;
 extern VALUE rb_eRuggedError;
@@ -18,6 +19,7 @@ extern VALUE rb_cRuggedConfig;
 extern VALUE rb_cRuggedBackend;
 extern VALUE rb_cRuggedRemote;
 extern VALUE rb_cRuggedCommit;
+extern VALUE rb_cRuggedDiff;
 extern VALUE rb_cRuggedTag;
 extern VALUE rb_cRuggedTree;
 extern VALUE rb_cRuggedReference;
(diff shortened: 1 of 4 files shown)


dometto:
Just wanted to add that similar support is being added to PyGit: https://github.com/libgit2/pygit2/pull/843/

dometto:
Ping. Are you accepting PRs like these at the moment?

ext/rugged/rugged_repo.c
```diff
@@ -2543,6 +2581,33 @@ static VALUE rb_git_repo_cherrypick_commit(int argc, VALUE *argv, VALUE self)
 	return rugged_index_new(rb_cRuggedIndex, self, index);
 }
 
+void rugged_parse_apply_options(git_apply_options *opts, git_apply_location_t *location, VALUE rb_options)
```
vmg: I see you're passing in the `git_apply_options` here, but it's not being actually filled by any of the fields of the `rb_options` hash.

dometto:
@vmg thanks for the review. The `git_apply_options` struct wasn't being filled as I hadn't yet implemented support for the callbacks, which were the only options. I've implemented the callbacks now (including tests), so please have a look!

`libgit2` determines whether a hunk/delta should be applied based on whether the relevant callback's return value `== 0` (apply, continue), `<0` (don't apply, continue) or `>0` (don't apply, abort). The ruby callbacks (blocks) must evaluate to `true`, `false`, and `nil` respectively to trigger the corresponding behavior. If there is a different standard for this, please let me know.

I've also changed the default `:location` to `:workdir`, as this conforms with the CLI usage of `git apply`. One downside to this, however, is that by default `apply` will trigger a `Rugged::OSError: could not stat 'filename': No such file or directory.` when run on a bare repository. `:index` would thus be the safest default. Or should we let the default location be determined by whether the repository is bare or not? (For example: in case it is bare, default to `:index`, otherwise default to `:workdir`?)

vmg:
> Or should we let the default location be determined by whether the repository is bare or not? (For example: in case it is bare, default to :index, otherwise default to :workdir?)

I think this would be a very nice UX feature for the library.

Code looks great btw :smile: 

@carlosmn can you give this a look please?

dometto:
> I think this would be a very nice UX feature for the library.
 
👍 Done!

> Code looks great btw 😄

Many thanks!

vmg:
This was a very neat PR. Thanks for your contribution!

dometto:
My pleasure! Do you perhaps have an approximate timeframe for the next release?

dometto:
@carlosmn I saw 0.28.2 was just released, but that doesn't include the changes in this PR. Any ETA for a new release containing this new feature?

dometto:
@vmg @carlosmn sorry to be so whiny, but could you let me know when I could expect a release that includes these changes? I'm asking because the next major release of https://github.com/gollum/gollum is hanging on this (as well as some unfinished business on our side), and it would be great to be able to plan our work a little. Many thanks!

PR #790, opened 2019-04-03
[Feature] Expose libgit2's git_diff_from_buffer in rugged through Rugged::Repository#diff_from_buffer(buffer)

bartkamphorst:
This PR exposes libgit2's `git_diff_from_buffer(git_diff **out, const char *content, size_t content_len)` [function](https://github.com/libgit2/libgit2/blob/master/src/diff_parse.c#L68-L107) in rugged by adding a class method `Rugged::Diff::from_buffer(buffer)`. This allows for creating Rugged::Diff objects from patches in String format. Please have a look at the code and let @dometto and myself know if this PR is on the right track.

BTW, one peculiar thing we noticed while writing the tests for this PR is that `git_diff_from_buffer` rejects patches that are created with `Rugged::Repository#diff(left, right, options={})` if either `left` or `right` is set to `nil`. According to [the test suite](https://github.com/libgit2/rugged/blob/master/test/diff_test.rb#L108-L157) this is allowed, but as I said, the resulting patches are rejected with error code `-1`. For example, the following creates three patches, the first two of which are rejected by `git_diff_from_buffer`, but the third works fine:
```ruby
repo = FixtureRepo.from_libgit2("attr")
patch1 = repo.diff(nil, "370fe9ec22").patch
patch2 = repo.diff("605812a", nil).patch
patch3 = repo.diff("605812a", "370fe9ec22").patch
puts patch1.inspect
puts patch2.inspect
puts patch3.inspect
```

Can anyone (@ethomson ?) help explain this behavior? Does it make sense or is this a bug?

### DIFF
--- a/ext/rugged/rugged_diff.c
+++ b/ext/rugged/rugged_diff.c
@@ -657,36 +657,9 @@ static VALUE rb_git_diff_sorted_icase_p(VALUE self)
 	return git_diff_is_sorted_icase(diff) ? Qtrue : Qfalse;
 }
 
-/*
- *  call-seq: Diff.from_buffer(buffer) -> Rugged::Diff object
- *  
- *  Where +buffer+ is a +String+.
- *  Returns A Rugged::Diff object
- */
-static VALUE rb_git_diff_from_buffer(VALUE self, VALUE rb_buffer)
-{
-  git_diff *diff = NULL;
-  const char *buffer;
-  size_t len;
-  VALUE rb_diff;
-  int error;
-
-  Check_Type(rb_buffer, T_STRING);
-  buffer = RSTRING_PTR(rb_buffer);
-  len = RSTRING_LEN(rb_buffer);
-
-  error = git_diff_from_buffer(&diff, buffer, len);
-  rugged_exception_check(error);
-
-  rb_diff = Data_Wrap_Struct(rb_cRuggedDiff, NULL, git_diff_free, diff);
-  return rb_diff;
-}
-
 void Init_rugged_diff(void)
 {
 	rb_cRuggedDiff = rb_define_class_under(rb_mRugged, "Diff", rb_cObject);
-
-	rb_define_singleton_method(rb_cRuggedDiff, "from_buffer", rb_git_diff_from_buffer, 1);
 
 	rb_define_method(rb_cRuggedDiff, "patch", rb_git_diff_patch, -1);
 	rb_define_method(rb_cRuggedDiff, "write_patch", rb_git_diff_write_patch, -1);
--- a/ext/rugged/rugged_diff.c
+++ b/ext/rugged/rugged_diff.c
@@ -657,9 +657,28 @@ static VALUE rb_git_diff_sorted_icase_p(VALUE self)
 	return git_diff_is_sorted_icase(diff) ? Qtrue : Qfalse;
 }
 
+static VALUE rb_git_diff_from_string(VALUE contents)
+{
+	git_diff *diff;
+	const char *buffer;
+	size_t len;
+	VALUE rb_diff;
+
+	buffer = StringValuePtr(contents);
+	len = strlen(buffer);
+
+	git_diff_from_buffer(&diff, buffer, len);
+
+	Data_Wrap_Struct(rb_cRuggedDiff, NULL, git_diff_free, diff);
+
+	return rb_diff;
+}
+
 void Init_rugged_diff(void)
 {
 	rb_cRuggedDiff = rb_define_class_under(rb_mRugged, "Diff", rb_cObject);
+
+	rb_define_singleton_method(rb_cRuggedDiff, "new_from_string", rb_git_diff_from_string, 1);
 
 	rb_define_method(rb_cRuggedDiff, "patch", rb_git_diff_patch, -1);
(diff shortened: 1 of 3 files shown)


ext/rugged/rugged_diff.c
```diff
@@ -657,10 +657,37 @@ static VALUE rb_git_diff_sorted_icase_p(VALUE self)
...
+	int error;
+
+  Check_Type(rb_buffer, T_STRING);
+  buffer = RSTRING_PTR(rb_buffer);
+  len = RSTRING_LEN(rb_buffer);
+
+  error = git_diff_from_buffer(&diff, buffer, len);
+  rugged_exception_check(error);
+  
+  rb_diff = Data_Wrap_Struct(rb_cRuggedDiff, NULL, git_diff_free, diff);
```
vmg: Note that all the other instances of new `diff` objects are created through `rugged_diff_new`, which wraps the diff _and_ sets a Repository owner for the object. I would encourage you to move add a "repository" argument to this method so we can store the repository object as the owner of the created diff -- in theory there should be no repository-owned data in the diff, but it's good that all `Diff` objects created by the library are consistently owned by a `Repository`.

vmg: ...Alternatively I think we could just make this a method on `Repository`, just like `Repository#diff`.

vmg:
Looking at the code more carefully, I'm not sure of why are we calling `rugged_set_owner` on these diff objects. @ethomson: do you know whether the lifetime of a `git_diff` in libgit2 is tied to a `git_repository`?

ext/rugged/rugged_diff.c
```diff
@@ -657,10 +657,37 @@ static VALUE rb_git_diff_sorted_icase_p(VALUE self)
...
+ */
+static VALUE rb_git_diff_from_buffer(VALUE self, VALUE rb_buffer)
+{
+	git_diff *diff = NULL;
+  const char *buffer;
+	size_t len;
+	VALUE rb_diff;
+	int error;
+
+  Check_Type(rb_buffer, T_STRING);
```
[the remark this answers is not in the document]
bartkamphorst: Fixed.

bartkamphorst:
> I'm not sure of why are we calling `rugged_set_owner` on these diff objects.

I was wondering about this as well.

ethomson:
> `repo.diff("605812a", nil)`

I don’t know enough about rugged to say, but yes, I would expect to be able to diff a commit against `nil` or some otherwise empty structure.

Otherwise I’m afraid I’m on holiday, so I don’t have time (or sufficient internet connectivity) to look at this closely for a while.

vmg:
@bartkamphorst: I'm afraid we're gonna have to wait until Ed is back from :ship: :swimming_man: :sun_with_face: before we can move forward with this.

@ethomson don't get sunburnt!!!!!

bartkamphorst:
@vmg OK, no problem. We'll pick this up when @ethomson is back. Thanks for your input so far! @ethomson Enjoy your time off!

ethomson:
> @ethomson: do you know whether the lifetime of a git_diff in libgit2 is tied to a git_repository?

Yes, I can confirm this.  The diff has [a reference to its owner](https://github.com/libgit2/libgit2/blob/master/src/diff_generate.c#L418).

ethomson:
> According to the test suite this is allowed, but as I said, the resulting patches are rejected with error code -1. 

What's the actual error message?  That should give some insight into the problem.  I don't really know, but I notice for example in the test suite that there are some odd settings to `context_lines` and `interhunk_lines`.  If rugged is setting those to 0 by default, then it's going to produce patches that can't be applied.  (You can create a patch with no context but you can't apply one).

ext/rugged/rugged_diff.c
bartkamphorst: > Alternatively I think we could just make this a method on Repository

Given that a Diff indeed needs an owner, I think this suggestion is probably the most intuitive solution. I'll implement `Rugged::Repository#diff_from_buffer(s)` instead.

bartkamphorst: @vmg Would you prefer to also move the tests to `repo_test.rb` or do I keep it under `diff_test.rb` as it is a test for creating a diff?

ext/rugged/rugged_repo.c
```diff
@@ -2710,6 +2710,29 @@ static VALUE rb_git_repo_cherrypick_commit(int argc, VALUE *argv, VALUE self)
...
 
+/*
+ *  call-seq: repo.diff_from_buffer(buffer) -> Rugged::Diff object
+ *  
+ *  Where +buffer+ is a +String+.
+ *  Returns A Rugged::Diff object
+ */
+static VALUE rb_git_diff_from_buffer(VALUE self, VALUE rb_buffer)
+{
+  git_diff *diff = NULL;
```
vmg: Code looks good but indentation is wrong again :sweat_smile:

bartkamphorst: Sorry about that. It looked the same in my instance of TextMate, but I should've checked more closely. Should be fixed.

ext/rugged/rugged_diff.c
bartkamphorst: @vmg Any thoughts about where to place the tests?

bartkamphorst:
Thanks @vmg for the quick review and helpful comments!

PR #792, opened 2019-04-15
Fix Windows builds

cfillion:
This pull request addresses two issues preventing rugged 0.28 from installing successfully on Windows:

- `pgroup` is not available on Windows (the alternative is `new_pgroup`)
- The mingw `libssh2` package, which is required for building rugged, is not installed out of the box on RubyInstaller with MSYS2


RoryO:
@carlosmn @tenderlove can we get eyes on this? currently windows users must install 0.27

RoryO:
HOLY BUCKETS _how are you that fast_

Issue #796, opened 2019-05-09
support ssh keys from memory


ccutrer:
libgit2 already supports ssh keys from memory (https://libgit2.org/libgit2/#HEAD/group/cred/git_cred_ssh_key_memory_new), just need to wrap that with a suitable ruby object

PR #797, opened 2019-05-09
implement SshKeyFromMemory

ccutrer:
fixes gh-796

libgit2 already supports it.

I wrote the tests for it, but was unable to get _any_ of the OnlineTests to work. I'm guessing I probably just don't fully understand what I'm supposed to put in the environment variables. I did test it for reals. And I also tested it with libgit2 not having libssh2 support (since ssh key from memory is an optional feature), and Rugged's extensions still compiles. I did not test it with a libgit2 with libssh2 support, but _not_ support for ssh key from memory (presumably a crypto backend other than openssl?).

### DIFF
--- a/test/online/fetch_test.rb
+++ b/test/online/fetch_test.rb
@@ -81,6 +81,14 @@ class OnlineFetchTest < Rugged::OnlineTestCase
       })
     end
 
+    def test_fetch_over_ssh_with_credentials_from_memory
+      @repo.remotes.create("origin", ENV['GITTEST_REMOTE_SSH_URL'])
+
+      @repo.fetch("origin", {
+          credentials: ssh_key_credential_from_memory
+      })
+    end
+
     def test_fetch_over_ssh_with_credentials_from_agent
       @repo.remotes.create("origin", ENV['GITTEST_REMOTE_SSH_URL'])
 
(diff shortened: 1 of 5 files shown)


ccutrer:
I've now tested with libssh2 built against gcrypt. when attempting to use in-memory ssh key without a public key (as you can with OpenSSL), you get `Rugged::SshError: Failed to authenticate SSH session: Unable to extract public key from private key.`. When I also provide the public key, I get `Rugged::SshError: Failed to authenticate SSH session: Unable to send userauth-publickey request`

ccutrer:
I'm pretty sure the previous error was because libgcrypt can't handle my private keys at all. When using a libssh2 _with_ openssl, but _without_ ssh memory credentials (???), I get `Rugged::InvalidError: this version of libgit2 was not built with ssh memory credentials.`. Which seems hugely descriptive to me of the exact problem!

test/online/fetch_test.rb
```diff
@@ -81,6 +81,14 @@ def test_fetch_over_ssh_with_credentials
       })
     end
 
+    def test_fetch_over_ssh_with_credentials_from_memory
+      @repo.remotes.create("origin", ENV['GITTEST_REMOTE_SSH_URL'])
+
+      @repo.fetch("origin", {
+          credentials: ssh_key_credential_from_memory
```
carlosmn, line 88: This will work as long as we keep compiling against libssh2 which itself is built aginst openssl, but otherwise we'll start seeing errors in test.

PR #798, opened 2019-05-24
fix default arguments for Commit#diff when nothing is passed

ccutrer:
this was broken by https://github.com/libgit2/rugged/commit/9bea4004984b9ed60741b4bdff5d32454ed76132 which neglected to handle the case where _no_ arguments are passed at all (not just nothing to diff against).

### DIFF
--- a/lib/rugged/commit.rb
+++ b/lib/rugged/commit.rb
@@ -20,10 +20,24 @@ module Rugged
 
     # Return a diff between this commit and its first parent or another commit or tree.
     #
+    # The commit is treated as the new side of the diff by default
+    #
     # See Rugged::Tree#diff for more details.
     def diff(*args)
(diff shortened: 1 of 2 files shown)


lib/rugged/commit.rb
```diff
@@ -20,10 +20,24 @@ def header_field?(field)
...
+      if other.is_a?(Hash)
+        opts = other
+        other = nil
+      end
+      opts ||= {}
+      # if other is not provided at all (as opposed to explicitly nil, or given)
+      # then diff against the prior commit
+      if args.empty? || args.first.is_a?(Hash)
+        other = parents.first
+        opts[:reverse] = !opts[:reverse] if other
```
carlosmn, line 38: Why are we checking `if other` here? If `self` is a root commit, `parents.first` will be `nil` but even in that case we want to reverse the order, or it will look the wrong way around.

PR #805, opened 2019-07-24
Use TypedData when wrapping library objects

carlosmn:
This lets us provide more information to the ruby VM about what our objects look
like including information about how much memory they use up.

### DIFF
--- a/ext/rugged/rugged_object.c
+++ b/ext/rugged/rugged_object.c
@@ -16,6 +16,8 @@ extern VALUE rb_cRuggedRepo;
 
 VALUE rb_cRuggedObject;
 
+const rb_data_type_t rugged_object_type;
+
 git_otype rugged_otype_get(VALUE self)
 {
 	git_otype type = GIT_OBJ_BAD;
@@ -74,7 +76,7 @@ int rugged_oid_get(git_oid *oid, git_repository *repo, VALUE p)
 	int error;
 
 	if (rb_obj_is_kind_of(p, rb_cRuggedObject)) {
-		Data_Get_Struct(p, git_object, object);
+		TypedData_Get_Struct(p, git_object, &rugged_object_type, object);
 		git_oid_cpy(oid, git_object_id(object));
 	} else {
 		Check_Type(p, T_STRING);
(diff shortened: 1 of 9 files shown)


ext/rugged/rugged_object.c
```diff
@@ -346,14 +393,14 @@ static VALUE rb_git_object_type_GET(VALUE self)
...
-	Data_Get_Struct(self, git_object, object);
+	TypedData_Get_Struct(self, git_object, &rugged_object_type, object);
 
 	return rugged_raw_read(git_object_owner(object), git_object_id(object));
 }
 
 void Init_rugged_object(void)
 {
-	rb_cRuggedObject = rb_define_class_under(rb_mRugged, "Object", rb_cObject);
+	rb_cRuggedObject = rb_define_class_under(rb_mRugged, "Object", rb_cData);
```
vmg: I see you've read the docs. :smile: It seems like inheriting `rb_cData` is a requirement, but could this possibly break code in the Ruby part of the library? :thinking:

ext/rugged/rugged_object.c
```diff
@@ -132,11 +134,56 @@ git_object *rugged_object_get(git_repository *repo, VALUE object_value, git_otyp
...
+		break;
+	default:
+		break;
+	}
+
+	return size;
+}
+
+const rb_data_type_t rugged_object_type = {
+	.wrap_struct_name = "Object",
```
vmg: This is a global identifier and doesn't need to be valid Ruby or C, so calling this `Rugged::Object` would be good: otherwise we're gonna see a lot of random "object" in our heap dumps.

ext/rugged/rugged_object.c
carlosmn: Based on

    object.c:4398:    rb_cData = rb_define_class("Data", rb_cObject);

it should be fine since we end up with the base object anyway. The specific "git object" classes already have a level of indirection between them and base ruby object, so I don't expect issues. We would still however roll this out somewhat slowly and I would expect both this test suite and our applications' test suites to notice if it breaks.

ext/rugged/rugged_object.c
carlosmn: Yep, that sounds good.

carlosmn:
Also cc @tenderlove as this came out of our discussion

I was looking at going deeper, but it might make more sense to attempt to apply these changes to our problem app and make sure we don't break anything. We can always come back and be more comprehensive.

In particular, we should be able to get libgit2 to expose actual object sizes since we already have that information in the object, just not exposed. Accurately counting how much memory a repository is hold on to is going to be harder though... including some caches which may not be trivial to count.

Issue #440, opened 2014-11-28
Proxy support for HTTP transport for `.clone_at` in rugged?


ghost:
Is there any chance to support HTTP-transport via proxy in `rugged` for `Rugged::Repository.clone_at('http://example.org/repo.git')`? There's an issue open at libgit2/libgit2#2555, but maybe that would be something which can / needs to be "fixed" in `rugged` itself.

Cheers,
dg

arthurschreiber:
From what I understand, http(s) connections that make use of `libcurl` will honor the `http_proxy` and `https_proxy` environment variable correctly. Unfortunately, there's no runtime support for setting this configuration yet - https://github.com/libgit2/libgit2/pull/3110 is supposed to resolve that, but it's still pretty much a work-in-progress.

To summarize: If you run on linux, use rugged 0.23 or later, and have the libgit2 bits compiled to use libcurl, proxies for http and https should work based on what is defined in the `http_proxy` or `https_proxy` environment variables.

synthead:
Has this been resolved with https://github.com/libgit2/rugged/pull/808?

PR #807, opened 2019-08-05
Add http proxy support

carbonin:
Add proxy support for clone and remote operations.

Hoping this will address #440 

Looking for input on the API (Should this live in an option?) and help with how I should write tests for this.

cc @fryguy @nicklamuro
Also @tenderlove saw you in some commits :wave:


carbonin:
This utilizes the non-libcurl proxy support added to libgit2 in https://github.com/libgit2/libgit2/pull/4870

PR #808, opened 2019-08-05
Add http proxy support

carbonin:
Add proxy support for clone and remote operations.
This utilizes the non-libcurl proxy support added to libgit2 in libgit2/libgit2#4870

Hoping this will address #440

Looking for input on the API (Should this live in an option?) and help with how I should write tests for this.

cc @Fryguy @NickLaMuro
Also @tenderlove saw you in some commits :wave:

### DIFF
--- a/ext/rugged/rugged_remote.c
+++ b/ext/rugged/rugged_remote.c
@@ -207,6 +207,18 @@ static void init_custom_headers(VALUE rb_options, git_strarray *custom_headers)
 	}
 }
 
+static void init_proxy_options(VALUE rb_options, git_proxy_options *proxy_options)
+{
+	if (NIL_P(rb_options)) return;
+
+	VALUE val = rb_hash_aref(rb_options, CSTR2SYM("proxy_url"));
+	if (!NIL_P(val)) {
+		Check_Type(val, T_STRING);
+		proxy_options->type = GIT_PROXY_SPECIFIED;
+		proxy_options->url = StringValueCStr(val);
+	}
+}
+
 static int parse_prune_type(VALUE rb_prune_type)
 {
 	if (rb_prune_type == Qtrue) {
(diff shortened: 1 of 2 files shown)


carbonin:
Sorry for the extra noise, but I force-pushed to my branch to fix up the documentation I wrote (https proxies are not supported by libgit2), but the PR didn't update ... not sure what that's about.

carbonin:
@carlosmn, not sure if you're the right person to ping here, but could you take a look at this?

Thanks!

jfine:
Would be great if this 🚢'd! We're using environment variable in the meantime.

Fryguy:
@carlosmn We could really use this PR merged if you could review.

synthead:
Could we get some :eyes: on this one?  It would be great to have this merged! :heart:

Magisus:
I'm also interested in seeing this go in, hoping to pick it up for use in [r10k](https://github.com/puppetlabs/r10k). We've been having trouble getting proxies working right since the libgit2 update dropping curl, and I think this might be the missing link for us, that's cleaner than what we used to do anyway.

ethomson:
also cc @pks-t

Magisus:
The behavior we were using before in r10k relied on libcurl checking environment variables for proxies by default. The new implementation in libgit no longer does this; instead you have to configure `fetch_opts.proxy_opts.type = GIT_PROXY_AUTO`.

We can probably use what this PR adds with some changes to our [proxy handling](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/git.rb#L175-L194), but I was curious if anyone else wanted to use environment variables to configure proxies? If so, it might be worth also adding that option. I might look into it, especially if this route doesn't wind up being feasible for us.

carlosmn:
Looks good thanks. Potentially we could see an `:auto` value in order to set to the `_AUTO` but that can be part of some later change.

Issue #819, opened 2019-09-13
Crash in rb_git_repo_file_each_status


ioquatix:
```
x_x > bundle exec teapot fetch --update
/Users/samuel/.rvm/gems/ruby-head/gems/teapot-3.1.0/lib/teapot/select.rb:45: warning: The last argument is used as the keyword parameter
/Users/samuel/.rvm/gems/ruby-head/gems/teapot-3.1.0/lib/teapot/configuration.rb:38: warning: for `initialize' defined here
/Users/samuel/.rvm/gems/ruby-head/gems/console-1.4.0/lib/console/filter.rb:143: warning: The last argument is used as the keyword parameter
/Users/samuel/.rvm/gems/ruby-head/gems/console-1.4.0/lib/console/terminal/logger.rb:79: warning: for `call' defined here
  0.1s: Processing clones platforms from http://github.com/kurocha/platforms...
/Users/samuel/.rvm/gems/ruby-head/gems/console-1.4.0/lib/console/filter.rb:143: warning: The last argument is used as the keyword parameter
/Users/samuel/.rvm/gems/ruby-head/gems/console-1.4.0/lib/console/terminal/logger.rb:79: warning: for `call' defined here
  0.1s: Updating package at path /Users/samuel/Documents/kurocha/async-network/teapot/packages/development/platforms...
/Users/samuel/.rvm/gems/ruby-head/gems/rugged-0.28.3.1/lib/rugged/repository.rb:95: [BUG] Unknown longjmp status 32712
ruby 2.7.0dev (2019-09-13T10:48:44Z trunk 7e0f56fb3d) [x86_64-darwin18]

-- Crash Report log information --------------------------------------------
   See Crash Report log file under the one of following:                    
     * ~/Library/Logs/DiagnosticReports                                     
     * /Library/Logs/DiagnosticReports                                      
   for more details.                                                        
Don't forget to include the above Crash Report log file in bug reports.     

-- Control frame information -----------------------------------------------
c:0017 p:---- s:0106 e:000105 CFUNC  :each_status
c:0016 p:0022 s:0102 e:000101 METHOD /Users/samuel/.rvm/gems/ruby-head/gems/rugged-0.28.3.1/lib/rugged/repository.rb:95 [FINISH]
c:0015 p:---- s:0096 e:000095 CFUNC  :each
c:0014 p:---- s:0093 e:000092 CFUNC  :any?
[...]

PR #820, opened 2019-09-13
Ensure exception is initialized, in case nentries is zero.

ioquatix:
Fixes https://github.com/libgit2/rugged/issues/819.


ioquatix:
@carlosmn do you mind merging this?

ioquatix:
The failure is due to a bug in the CI build:

Screen Shot 2019-11-20 at 1 15 40 PM

PR #825, opened 2019-12-09
Adding Windows CI, fixing tests on Windows, test robustness

RoryO:
~draft~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,9 @@ jobs:
     steps:
     - uses: actions/checkout@master
 
+    - name: Update submodules
+      run: git submodule init && git submodule update
+
     - name: Install Linux packages
       if: runner.os == 'Linux'
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,39 +12,32 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.3.7', '2.4.6', '2.5.5', '2.6.2' ]
+        ruby: [ '~> 2.4', '~> 2.5', '~> 2.6' ]
         os: [ ubuntu-18.04, macOS-10.14 ]
 
     runs-on: ${{ matrix.os }}
 
-    name: Ruby ${{ matrix.ruby }} on ${{ matrix.os }}
+    name: ${{ matrix.ruby }} on ${{ matrix.os }}
     steps:
     - uses: actions/checkout@master
-    - name: update submodule
-      run: git submodule update --init
+
     - name: Install Linux packages
       if: runner.os == 'Linux'
       run: |
         sudo apt update
         sudo apt install -y cmake libssh2-1-dev openssh-client openssh-server
+
     - name: Install macOS packages
       if: runner.os == 'macOS'
       run: ./vendor/libgit2/azure-pipelines/setup-osx.sh
-    - name: Set up Ruby on Linux
-      if: runner.os == 'Linux'
+
+    - name: Set up Ruby
       uses: actions/setup-ruby@v1
       with:
         version: ${{ matrix.ruby }}
-    - name: Set up Ruby on macOS
-      if: runner.os == 'macOS'
-      run: |
-        brew install rbenv
-        rbenv install ${{ matrix.ruby }}
-        rbenv local ${{ matrix.ruby }}
+
     - name: run build
       run: |
-        if [ -x rbenv ]; then eval "$(rbenv init -)"; fi
-        ruby --version
         gem install bundler
         bundle install --path vendor
         ./script/travisbuild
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -12,39 +12,35 @@ jobs:
     strategy:
```
RoryO: Updating and simplifying this. Dropping 2.3 as it's EOL and not available in the CI images.

test/index_test.rb
```diff
@@ -138,11 +138,13 @@ def test_raises_when_writing_invalid_entries
 
   def test_can_write_index
```
RoryO: This was an interesting test failure. What was happening here is it accidentally working. The procedure adds two new index entries. Originally the OIDs are identical. Then, when checking for the order of insertion we sorted by the OID. The stability of this sort depends on too many factors, and was in fact in a different order on Windows. Quick fix is create the second index entry with an OID higher than the first.

test/repo_test.rb
```diff
@@ -554,12 +550,15 @@ def test_clone_bare
   end
 
   def test_clone_with_transfer_progress_callback
+    skip 'Callback does not work with filesystem clone on Windows' if Gem.win_platform?
```
RoryO: There may be some sort of difference with Windows where callbacks will not trigger unless a buffer hits a certain size. Other clone tests pass, so we know that cloning actually works, it's only the callback which doesn't trigger.

test/repo_test.rb
```diff
@@ -804,6 +802,11 @@ def verify_subtrees
   end
 
   def test_checkout_tree_with_commit
+    # the test repo has an unclean status. apparently libgit2 on *nix does not
```
RoryO: Another interesting thing I don't know how to resolve. The libgit2 test repo has modifications initially, confirmedon linux and windows. Apparently on *nix checking out another pointer isn't an issue here, on Windows it is for some reason.

test/test_helper.rb
```diff
@@ -169,3 +166,9 @@ def ssh_key_credential_from_agent
     end
   end
 end
+
+# Automatically clean up created fixture repos after the whole suite
+# there are race conditions on Windows where the original test run
+# did not release handles to the temporary directories.
+# waiting until all tests finish works successfully
+Minitest.after_run { Rugged::TestCase::FixtureRepo.teardown }
```
RoryO: The largest issue I kept running into with these tests is something about the Windows libgit2 was leaving open file handles to the temporary directories created. It appears the handles close when the test suite finishes. Therefore I think the best thing to do is clean up everything at the end of the suite instead of inline.

test/repo_test.rb
```diff
@@ -524,11 +518,13 @@ def test_init_with_wrong_argument
 class RepositoryCloneTest < Rugged::TestCase
   def setup
     @tmppath = Dir.mktmpdir
-    @source_path = "file://" + File.join(Rugged::TestCase::TEST_DIR, 'fixtures', 'testrepo.git')
-  end
+    Rugged::TestCase::FixtureRepo.ensure_cleanup @tmppath
```
RoryO: Making all temporary directory removal go through the same procedure for working around the open file handle issue mentioned later.

rugged.gemspec
```diff
@@ -31,5 +31,8 @@ desc
   s.add_development_dependency "rake-compiler", ">= 0.9.0"
   s.add_development_dependency "pry"
   s.add_development_dependency "minitest", "~> 5.0"
+  s.add_development_dependency "minitest-reporters"
```
RoryO: added for junit xml output. github actions does not support this at this time.

test/online/fetch_test.rb
```diff
@@ -54,7 +54,13 @@ def test_fetch_over_https_with_certificate_callback_fail
         })
       end
 
-      assert_equal "user rejected certificate for github.com", exception.message
+      response_message = if Gem.win_platform?
```
RoryO: the windows response is actually more correct to the scenario. strange how we get different messages for the same action.

test/remote_test.rb
```diff
@@ -124,6 +124,7 @@ def test_remote_lookup_invalid
 
 class RemotePushTest < Rugged::TestCase
   def setup
+    skip 'local files and file:// protocol handled inconsistently with libgit2 on windows' if Gem.win_platform?
```
RoryO: I'm already several layers of yak shaving doing what I originally wanted. This is worthy of someone looking into the libgit2 source and see what's up there. Some situations `file://` works and is necessary, others it's rejected by the kernel with an invalid path name.

RoryO:
Looks like linux builds fail for reasons that are not our own. Appears github is transitioning ubuntu repos from ubuntu.com to microsoft.com

https://github.com/libgit2/rugged/runs/346774249#step:4:61

RoryO:
Pinging @carlosmn @tenderlove for visibility

RoryO:
@carlosmn @tenderlove beep beep

RoryO:
@carlosmn @tenderlove hello

PR #826, opened 2019-12-10
Bump libgit2 to 6777db8e834

carlosmn:
This includes security fixes around NTFS alternate data streams.


carlosmn:
We're just gonna continue ignoring the 2.3 rubies because they're way too old to support considering chunks of our CI and ruby environment don't want to either.

PR #827, opened 2019-12-10
Update libgit2 to v0.27.10

carlosmn:


carlosmn:
The errors come from trying to test on too old a version of ruby so we can continue.

PR #828, opened 2019-12-10
Bump libgit2 to v0.28.4

carlosmn:


carlosmn:
The error comes from 2.3.3 on maOS which we don't really care about, we can continue.

Issue #818, opened 2019-09-12
Release schedule?


dometto:
Sorry to bother everyone about this once again, but it would be great to know when you're planning to release a version of the gem that includes commits 3f614b1be7314f037f5d7d6b9a45a5b5e106862f  to db63415b3f01606e850dceea2554e8e8cf240265 from `master`. The next major release of [gollum](https://github.com/gollum/gollum) is hanging on this (we need the `rugged` update to be able to revert files). For `gollum` and other packages that depend on it, it's the last thing that's keeping us from ditching `grit`. And we would really, really like to ditch grit. :)

bartkamphorst:
@carlosmn @vmg Seconding this request for a new rugged release. Thanks in advance for making the time!

dometto:
Another reminder of this. Is anything holding up a minor version bump?

bartkamphorst:
Still looking for anyone willing and able to release another gem version of rugged. @arthurschreiber Could you help out?

bartkamphorst:
Or perhaps @stanhu could jump in for a new release?

stanhu:
I'm not actually involved with cutting new releases. @carlosmn Can you help here? Thanks!

dometto:
@tenderlove @carlosmn the two (or four, if you count the bumps) that were released four days ago still don't seem to contain the relevant changes. :-( Is there any reason for not including them?

bartkamphorst:
I can confirm that [Repository#diff_from_buffer(buffer)](https://github.com/libgit2/rugged/pull/790) and [Repository#apply(diff, options = {})](https://github.com/libgit2/rugged/pull/788) have not made it into the latest 0.28.4.1 release, despite having been merged into `master` in April. @carlosmn @vmg @tenderlove Could you please shine a light on why this is? Many thanks!

mvz:
It looks like 0.28.4.1 and 0.27.10.1 were released from the `maint/0.28` and `maint/0.27` branches, respectively. The `maint/0.28` branch was branched off master in February. This means that any pull requests merged into master after that time won't have been included.

dometto:
@mvz thanks for explaining!

@carlosmn @vmg: is there anything that speaks against doing a new release from master? Or should we open a PR with the relevant changes to the `maint/X` branches?

dometto:
libgit2 `v0.99` has been released. Is a corresponding rugged release planned? If so, could you please include the changes to rugged `master`? Many thanks!

PR #829, opened 2020-01-16
Merge master into maintenance branch and bump version

dometto:
Please see #818. New versions have been created off the `maint/v0.28` branch (and its counterpart for `v0.27`), causing new features and fixes in `master` from the past year to be absent in new releases. This PR merges `master` into `maint/v0.28` and bumps the `rugged` version. It would be great if this could be merged, and a new release be made on the basis of it.

cc @tenderlove @vmg @carlosmn

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,50 @@
+name: Rugged CI
+
+on:
+  pull_request:
+  push:
+    branches:
+    - master
+    - maint/*
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [ '2.3.7', '2.4.6', '2.5.5', '2.6.3' ]
+        os: [ ubuntu-18.04, macOS-10.14 ]
+
+    runs-on: ${{ matrix.os }}
+
+    name: Ruby ${{ matrix.ruby }} on ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@master
+    - name: update submodule
+      run: git submodule update --init
+    - name: Install Linux packages
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt update
+        sudo apt install -y cmake libssh2-1-dev openssh-client openssh-server
+    - name: Install macOS packages
+      if: runner.os == 'macOS'
+      run: ./vendor/libgit2/azure-pipelines/setup-osx.sh
+    - name: Set up Ruby on Linux
+      if: runner.os == 'Linux'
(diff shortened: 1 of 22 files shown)


dometto:
Test failure seems to be due to a failure to setup ruby 2.3.7 on Linux on the CI.

ethomson:
> New versions have been created off the maint/v0.28 branch (and its counterpart for v0.27), causing new features and fixes in master from the past year to be absent in new releases.

v0.28 and v0.27 are maintenance branches, so they only get bugfixes.  What you've described is the appropriate behavior for them - they shouldn't be getting entirely new features from master.

> This PR merges master into maint/v0.28 and bumps the rugged version. It would be great if this could be merged, and a new release be made on the basis of it.

I'd encourage you instead to create a PR against master that revs the version to 0.29, as a new release branch.  I think that makes sense as the path forward for master instead of trying to backport new features into maintenance branches.

dometto:
@ethomson thanks for the switft reply. Happy to do so, but would that mean we would have to wait until `libgit2 v.029` is released? If so, can you give an approximate indication of when that would be?

ethomson:
Well, that is a good question.  I'm sorry, I hadn't realized that rugged tracked libgit2's releases in that way -- I'm not very familiar with rugged since I'm not much of a rubyist.  Though it does make sense.

libgit2 is preparing for an 0.99.  We have - in my mind - two issues left that I'd like to tackle before we do that.  We'll use this as a bugfix/stabilization release leading up to 1.0.

It might make sense for rugged to do an 0.99 prerelease instead of an 0.29?  That way it's a) pulling in the latest libgit2 bits, but not b) doing a full release ahead of the libgit2 release.

Maybe it would be called 0.29 prerelease, I'm not sure what makes the most sense here.

dometto:
> It might make sense for rugged to do an 0.99 prerelease instead of an 0.29? That way it's a) pulling in the latest libgit2 bits, but not b) doing a full release ahead of the libgit2 release.

If the rugged maintainers are prepared to do such a prerelease, that would be great! I'd happy to open a PR that pulls in the relevant bits of libgit2 and is based on rugged `master`. @ethomson what commit in libgit2 would you recommend pulling in for this? The current HEAD, i.e. a76348ee88c2ab4f079491f0a3810a12592b95cb?

dometto:
I see libgit2 v0.99 has just been released. @carlosmn what are the chances of incorporating the changes to rugged `master` in the (I assume) upcoming corresponding release of rugged?

PR #831, opened 2020-02-04
Adds pkg-config to the brew install command

san983:
Since `pkg-config` is required (and already mentioned in the doc. I'm adding it to the `brew install` snippet


san983:
Closing since https://github.com/libgit2/rugged/pull/845 addressed it 👍

PR #835, opened 2020-03-09
Bump libgit2 to 0.99.0

carlosmn:


Fryguy:
Is this right?  This bumped the rugged gem itself to 0.99.0 as well as the libgit2 reference...seems it was released on rubygems.org as well.

PR #838, opened 2020-04-03
Update libgit2 to v0.28.5

carlosmn:


carlosmn:
The CI error is due to ruby 2.3.3 not being accessible in the macOS build, and I feel confident to continue.

PR #840, opened 2020-04-07
Remove ruby 2.7 warnings

HParker:
Ruby 2.7 is raising a bunch of,
`warning: Using the last argument as keyword parameters is deprecated`
warnings. we can resolve these these by using ** when we intend that last argument to be kwargs.

### DIFF
--- a/lib/rugged/commit.rb
+++ b/lib/rugged/commit.rb
@@ -5,7 +5,6 @@
 
 module Rugged
   class Commit
-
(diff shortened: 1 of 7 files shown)


carlosmn:
Cool, looks like this works in all versions we care about.

jwark:
Hello :) 
Just wondering when this change is likely to be released?
Couldn't see anything about a release schedule, apologies if I have missed it. 
Thanks!

PR #841, opened 2020-04-08
Fix assigned but unused variable warnings in test

HParker:


HParker:
The failure seems to be a transient network issue.

carlosmn:
Thanks.

PR #842, opened 2020-04-08
config: test that we iterate over string keys

carlosmn:
We identified that #803 has changed the `Config#each_key` yielded values from strings to single-value arrays. This is not what was intended and breaks code that wants a string.

I'm opening this up with just the test to see if older ruby versions did treat a single-element array here as a single value, or if we really messed up.


carlosmn:
This is even broken in 2.3 so we clearly messed this up.

carlosmn:
Fixed in #839 this was mostly to check whether it was ever right via ci.

PR #845, opened 2020-04-24
Add more help on installing pkg-config on OSX

IanVaughan:
I ran this to get the required packages, maybe others will need to also?


carlosmn:
Sure, makes sense since we mention it in the line above anyway.

PR #847, opened 2020-06-10
Upgrade libgit2 to v1.0.1

pks-t:
There's several bugfixes released in v1.0.1, so Rugged should upgrade and probably pull a release itself. The most important fix is that cloning from Google Source now works again.


carlosmn:
Ta.

PR #852, opened 2020-08-19
Introduce Rugged::Submodule.valid_name?

HParker:
This allows checking if a submodule has a valid name and isn't trying to escape the repository.

### DIFF
--- a/ext/rugged/rugged_submodule.c
+++ b/ext/rugged/rugged_submodule.c
@@ -778,6 +778,26 @@ static VALUE rb_git_submodule_finalize_add(VALUE self)
 	return self;
 }
 
+/*
+ *  call-seq:
+ *    Submdule.valid_name? -> bool
+ *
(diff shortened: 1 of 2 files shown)


HParker:
@carlosmn I only get that warning when my `libgit2` submodule is behind. It seems that the `git_submodule_name_is_valid` method is available in libgit2 already.

https://github.com/libgit2/libgit2/blob/2376cd26226771dcf8ef5dfd04c83a50c50bf3d4/src/submodule.h#L162

Maybe I am confused about how the submodule here works, but it does seem that that method does exist in libgit2 already.

carlosmn:
> that method does exist in libgit2 already.

It does exist in libgit2, but it's not part of the public API, which is in the `src/` directory instead of `include/git2/`

HParker:
Thanks! I'll close this for now and reopen if I find a libgit2 API that makes sense. Thanks!

PR #854, opened 2020-09-01
fixed namespacing issue in example code

gilesbowkett:
Hi, this is just a super simple documentation fix. The README refers to a namespace/module which no longer exists, so I corrected it. `Rugged::InMemory::Backend` is now just `Rugged::Backend`. Thanks for your time!

### DIFF
--- a/README.md
+++ b/README.md
@@ -577,7 +577,7 @@ You can store bare repositories in alternative backends instead of storing on di
 `redbadger/rugged-redis` for an example of how a rugged backend works).
 
 ```ruby
-a_backend = Rugged::InMemory::Backend.new(opt1: 'setting', opt2: 'setting')
+a_backend = Rugged::Backend.new(opt1: 'setting', opt2: 'setting')
 
 repo = Rugged::Repository.init_at('repo_name', :bare, backend: a_backend)
 


carlosmn:
Thanks but this isn't right either. `Rugged::Backend` is a class without any actual implementation of anything. It also doesn't accept any arguments so that line is going to fail to run either way.

As far as I can tell, this is meant to be an example of *some* backend, not one that exists or existed in the codebase so perhaps the solution here is to use a namespace which is obviously not one that exists.

gilesbowkett:
yeah, it seems like a hook that was made for the specific example repo that the README uses, [rugged-redis](https://github.com/redbadger/rugged-redis). it's reasonable to assume that this worked at some point but I don't know if it still works today.

the rugged-redis README says that it uses a `Rugged::Backend` subclass, but as far as I can tell [the actual code is in C](https://github.com/redbadger/rugged-redis/tree/master/ext/rugged/redis). it has the general look and feel of a Ruby extension written in C, but I couldn't say for sure if it's still a `Rugged::Backend` subclass or not.

I didn't totally understand this part:

> perhaps the solution here is to use a namespace which is obviously not one that exists

as in, use example code in the README like `Rugged::Backend::FooBar`?

carlosmn:
> as in, use example code in the README like `Rugged::Backend::FooBar`?

No, one like `ExternalProject::RuggedBackend` which indicates it's not in this project.

gilesbowkett:
well, looks like 0327e217b56d24f10d32359da7e482534d2a3f5d resolved the ambiguity, so I'm closing this.

PR #857, opened 2020-09-18
Expose 'recursion_limit' option for merges

wfchandler:
libgit2 has a `recursion_limit` option in `git_merge_options` to
control the potential cost of recursive merging, but this is not
currently exposed in rugged.

This change makes `recursion_limit` available for use in rugged.


carlosmn:
Thanks. This would also need to be added to the list of supported options in the documentation for `rb_git_tree_merge` a bit further down the file.

scutuatua-crypto (approved):

PR #862, opened 2020-11-18
Its 2020

Planebagels:
and because it is 2020 the README just needs a small change.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License
 
-Copyright (c) 2016 GitHub, Inc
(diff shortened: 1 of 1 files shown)


Planebagels:
oops. I didnt mean README I meant the Liscense

seanfcarroll (changes requested):
Now it's 2021 😄

carlosmn:
The year rolling over doesn't mean a copyright notice has to increase. It would only do so if there have been changes in that year. It would also not *replace* the existing copyright notice as the copyrights from previous years would be just as valid.

PR #891, opened 2021-10-22
Bump version to 1.3.0

FnControlOption:


carlosmn:
It looks like maybe something changed in the CI setup for libgit2 that means we can't access the ssh stuff right now on macOS.

PR #894, opened 2021-12-19
ci: use ruby/setup-ruby for both builds

carlosmn:
Currently macOS is using rvm as the previous `actions/setup-ruby` did not work
there at the time were building this. This change should speed up and unify the
versions we run for our tests.


carlosmn:
macOS builds are down to ~8m from ~15m when we were installing by hand so that's nice.

PR #903, opened 2022-02-11
Allow setting pb_parallelism for remote push

bryanhelms:
We needed the option to provide a `pb_parallelism` value to libgit2 when pushing a remote, so this PR adds that ability.


bryanhelms:
I was able to confirm this works with a local repro, FYI. Let me know if there need to be any additional tests attached or anything.

PR #919, opened 2022-05-14
Update README.md

mslinn:
Added Debian/Ubuntu installation instructions, and a few subheads so readers can focus on the sections of interest to them.


carlosmn:
Tanks, I just added a note about when you do need the library installed on the system to hopefully avoid some confusion there.

Issue #926, opened 2022-07-14
No documentation on how to build rugged v1.4 with ssh support


agrare:
Hello!  I've noticed that starting with rugged `v1.4.2` the SSH feature is not enabled even with `libssh2` installed and available.  Installing rugged v1.3* in the same environment has SSH feature enabled as expected.

The difference appears to have started as of libgit2 v1.4.0 due to the change to `USE_SSH` now defaulting to off: https://github.com/libgit2/libgit2/pull/6112/commits/6c00fcb767b38c1902c0b1ca15350e7c5766e23d#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR43

This change doesn't appear to be in the libgit2 changelog or release notes and causes confusion (at least to me) as to why the SSH feature was suddenly disabled.

The rugged README points to the libgit2 README for SSH support: `If you want to build Rugged with HTTPS and SSH support, check out the list of optional libgit2 dependencies` but there is no indication there either that in addition to `LibSSH2 to enable the SSH transport` you have to pass an additional build flag.

Running `CMAKE_FLAGS='-DUSE_SSH=ON' gem install rugged` when libssh2 is installed results in the SSH feature being enabled.  It would be helpful if this were documented since otherwise the only way to discover that this is required is by digging into the libgit2 source.

A Build Option like `--with-ssh` might be helpful as well to simplify installing with bundler, e.g. `bundle config build.rugged --with-ssh`

PR #927, opened 2022-07-14
Add documentation on installing rugged with SSH

agrare:
As of libgit2 v1.4.0 USE_SSH defaults to off, this results in a behavior change from the end-user point-of-view in that rugged no longer has the SSH feature enabled even if libssh2 is installed where this worked on rugged v1.3.

Document the CMAKE_FLAGS needed to install rugged v1.4 with the SSH feature enabled.

https://github.com/libgit2/rugged/issues/926


agrare:
Replaced by https://github.com/libgit2/rugged/pull/928

PR #928, opened 2022-07-14
Add a --with-ssh build flag to set USE_SSH=ON

agrare:
Make it simpler to install rugged with SSH support via bundler by adding a `--with-ssh` build option which can be set by `bundle config build.rugged`

https://github.com/libgit2/rugged/issues/926


carlosmn:
Thanks. Ideally this would be set to automatic but it doesn't seem to have that option.

PR #948, opened 2023-01-21
Adds Ruby 3.2 to the CI matrix.

petergoldstein:
Runs green on my fork.

Please note that when making this change I initially upgraded the ubuntu-18.04 to a more recent ubuntu, but this causes all of the ubuntu jobs to fail.  There are ~7 failed specs, all of which look OpenSSL related, so I suspect a change in OpenSSL version causes them to fail.  This may require additional investigation.


carlosmn:
Thanks.

PR #950, opened 2023-03-02
fix: search at first system header for git2 then vendor's one

majioa:
- [x] ! search at first system header for git2 then vendor's one. please refer to [patch](https://git.altlinux.org/gears/g/gem-rugged.git?p=gem-rugged.git;a=blob;f=system_git2.patch;h=c6bd3b1e899522250c1fcfefdec51cb210531c6b;hb=f63322fa19292a2968e883e7870acac74aeff3e8)

### DIFF
--- a/ext/rugged/extconf.rb
+++ b/ext/rugged/extconf.rb
@@ -70,7 +70,10 @@ if arg_config("--use-system-libraries", !!ENV['RUGGED_USE_SYSTEM_LIBRARIES'])
 
   major = minor = nil
 
-  File.readlines(File.join(LIBGIT2_DIR, "include", "git2", "version.h")).each do |line|
+  git2_include_dir = Dir.glob("#{RbConfig::CONFIG["includedir"]}/**/git2").first ||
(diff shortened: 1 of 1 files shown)


carlosmn:
Is there a comment on what this is doing or why? The patch referenced is just the diff.

majioa:
In case the package libgit2-devel/dev is installed, the headers is picking up from them.

carlosmn:
That seems to be the opposite of what path is trying to do. But still, that doesn't tell me what this is trying to fix by searching in both when the options are for rugged to try to pick up either a system-installed libgit2 with the right version, OR use the vendored version.

If you ask it to use the vendred version, then it should use everything from the vendored libgit2. If you ask it to use the system version, then everything should come from the system version.

The logic you're modyfing is for the `--use system libraries" logic. If there is a libgit2 installed then it should only look at it. Looking at it now, it does look like it could buggy and it might be at the vendored header instead of the system header that was found. But then that's what needs to get fixed.

scutuatua-crypto (approved):

PR #956, opened 2023-03-23
add Credentials::SshKeyMemory (off v1.5.0.1)

tjk:
GitHub is not letting me create this PR against v1.5.0.1 which is the tag I branched off of...

---

Edit 1: looks like need to ensure libgit2 is compiled with `GIT_SSH_MEMORY_CREDENTIALS` defined... 

Edit 2: this may just be because it couldn't find libssh2 shared libs...

```
//Have library /usr/local/lib/libssh2.dylib
HAVE_LIBSSH2_MEMORY_CREDENTIALS:INTERNAL=1
```

Edit 3: ^ correct, this feature is disabled if libssh2 shared libs are not found (was previously building static). Not sure how to enforce or check in rugged... looking for guidance but at least for my use case this works.

### DIFF
--- a/ext/rugged/rugged_blame.c
+++ b/ext/rugged/rugged_blame.c
@@ -264,6 +264,8 @@ static VALUE rb_git_blame_each(VALUE self)
 void Init_rugged_blame(void)
 {
 	rb_cRuggedBlame = rb_define_class_under(rb_mRugged, "Blame", rb_cObject);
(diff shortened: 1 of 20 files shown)


tjk:
Need to track why but I actually think this is the cause of https://github.com/libssh2/libssh2/issues/1107... looking to see if it's something obvious.

scutuatua-crypto (approved):

PR #964, opened 2023-07-18
Update to 1.7.1

carlosmn:
Unfortunately we segfault but I'm not sure if this is something in rugged or libgit2 or elsewhere. We still segfault even without the custom allocator setup so presumably it's not something inside rugged.

Deeper testing would have to go through building a custom ruby that uses the system allocator so valgrind or gdb can see what is going on. Running a single test doesn't trigger this so it seems to be something larger, and it fails with a NULL pointer dereference somewhere deep in fileutils during the helper, which is quite odd.

### DIFF
--- a/ext/rugged/rugged_allocator.c
+++ b/ext/rugged/rugged_allocator.c
@@ -13,56 +13,9 @@ static void *rugged_gmalloc(size_t n, const char *file, int line)
 	return xmalloc(n);
 }
 
-static void *rugged_gcalloc(size_t nelem, size_t elsize, const char *file, int line)
-{
-	return xcalloc(nelem, elsize);
-}
-
-static char *rugged_gstrdup(const char *str, const char *file, int line)
-{
-	return ruby_strdup(str);
-}
-
-static char *rugged_gstrndup(const char *str, size_t n, const char *file, int line)
-{
-	size_t len;
-	char *newstr;
-
-	len = strnlen(str, n);
-	if (len < n)
-		n = len;
-
-	newstr = xmalloc(n+1);
-	memcpy(newstr, str, n);
-	newstr[n] = '\0';
-
-	return newstr;
-}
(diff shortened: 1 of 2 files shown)


dvzrv:
Hi! Since we're currently blocked by ruby-rugged in our [libgit2 1.7.0 rebuild](https://archlinux.org/todo/libgit2-17/) on Arch Linux:
Do you know if applying https://github.com/libgit2/libgit2/commit/9d4c550564ee254dda9e2620c4c1e32ebb529728 on top of libgit2 1.7.0 is enough to make your changes to rugged work?

carlosmn:
@dvzrv it was enough to get it to work on my machine, and updating to 1.7.1 it does work, which shouldn't have anything else regarding segfaults.

--- 

Unfortunately now that 1.7.1 is out and I'm trying to update it, all of the macOS stuff fails but I don't know how much even has to do with libgit2 vs how it's built on macOS. And I'm not where my mac machine is so I can't dig deeper at the moment.

ethomson:
@carlosmn I have a Mac, how can I help?

carlosmn:
I'm now close to my mac, but it's packed up somewhere... basically in CI on macOS all the ssh tests fail for macOS and I've no idea if it's a macOS issue or if there's something that changed with GitHub or what.

carlosmn:
The problems I had locally ended up being just because we don't actually kill sshd so it was looking at the wrong checksum. The tests run fine on my now-up-to-date machine that's a patch version away from the runner.

So it looks like the issue might be something related to how the runners work on Actions and I've no idea how to debug that.

carlosmn:
In the logs from sshd it looks like there's too many auth failures, and it's tried `none` and `publickey` repeatedly. I don't know if we really now need to use a different scheme from the deprecated RSA or if it's something else.

PR #967, opened 2023-08-25
test: remove obsolete `tee` to global config

carlosmn:
We use to modify the global sshd but we haven't done that for a few years. We still did have this `tee` to allow password logins that we don't need anymore. Remove this single instance of a `sudo`.


carlosmn:
The mac errors are the network ones that we seemingly get regardless of the rugged version.

PR #968, opened 2023-08-29
[DNM] debugging around actiosn on macOS

carlosmn:


carlosmn:
Well it looks like giving it a bit of time I can do a `ssh-keyscan` and have it show that we are talking locally. But there's still some issue getting a response... I'm wondering if this is a v4 vs v6 thing inside the runner or something... unlikely, but...

carlosmn:
Well now it's back to failing differently, so it is able to connect but not very well. This is extra confusing because libgit2 uses the same setup code but it's fine there.

PR #971, opened 2023-09-29
settings: Expose GIT_OPT_SET_OWNER_VALIDATION option

alexjfisher:
By default, `libgit2` will throw an error if you try to open a repository not owned by the current user. An option to disable this validation is available, but wasn't being exposed in `rugged`.

This commit exposes fetching and setting this option via a new `owner_validation` setting.

(My use case is using `rugged` in a ruby language AWS lambda function connected to EFS. In this scenario, the EFS access point is configured with a POSIX user ID and group ID `used for all file system operations using this access point`, but the lambda's execution environment's uid is not known upfront and won't match).


carlosmn:
Thanks.

PR #985, opened 2025-01-15
ext: Fix the build with system libgit2 1.9.0

petefwalter:
Update version.h parsing regexes for 1.9.0 and fix compilation test by including git2.h instead of directly including git2/version.h, which is no longer supported.


carlosmn:
Thanks.

PR #991, opened 2025-09-08
Check for strings before attempting to xfree them

### LINKED ISSUE #859: segfault in check_connection
We're seeing what appears to be a segfault in the `check_connection` method, i.e. `rb_git_remote_check_connection`. The error is `check_connection munmap_chunk(): invalid pointer`. Near as I can tell this happens when you try free something that wasn't malloc'ed.

https://github.com/libgit2/rugged/blob/master/ext/rugged/rugged_remote.c#L503-L539

Looking at that can we confirm that the `xfree` is only freeing things that were xmalloced? Also, should the code only call `git_remote_disconnect` if the connection actually happened?

Unfortunately, we have not been able to duplicate this in a standalone script. It seems to happen when the network goes down after previously successful connections.

Here is the actual code that triggered it:

https://github.com/ManageIQ/manageiq/pull/20645/files#diff-d54d482e470404136b4b893c60065b5d743c912652d58d69820d86604e8a8983R261-R268

More information:

https://bugzilla.redhat.com/show_bug.cgi?id=1835226

cc @NickLaMuro, @d-m-u

djberg96:
The `rugged_rb_ary_to_strarray` function uses `xmalloc` to allocate memory for `str_array->strings`, but there are several cases where this memory might not be allocated:

If `rb_array` is `NIL_P`, the function returns early with `strings = NULL`. If there's an empty array, the count would be 0. However, in the cleanup sections, the code unconditionally calls `xfree(custom_headers.strings)` without checking if it's NULL. This can cause "munmap_chunk(): invalid pointer" errors.

Should fix https://github.com/libgit2/rugged/issues/859

### DIFF
--- a/ext/rugged/rugged.c
+++ b/ext/rugged/rugged.c
@@ -438,7 +438,8 @@ void rugged_rb_ary_to_strarray(VALUE rb_array, git_strarray *str_array)
 	if (TYPE(rb_array) == T_STRING) {
 		const char *s = StringValueCStr(rb_array);
 		str_array->count = 1;
-		str_array->strings = xmalloc(sizeof(char *));
+		/* use malloc so git_strarray_dispose/free can clean up */
+		str_array->strings = malloc(sizeof(char *));
 		/* duplicate the string so it can be freed later */
 		str_array->strings[0] = strdup(s);
 		return;
@@ -450,7 +451,8 @@ void rugged_rb_ary_to_strarray(VALUE rb_array, git_strarray *str_array)
 		Check_Type(rb_ary_entry(rb_array, i), T_STRING);
 
 	str_array->count = RARRAY_LEN(rb_array);
-	str_array->strings = xmalloc(str_array->count * sizeof(char *));
+	/* allocate with malloc rather than xmalloc to allow free() to work */
+	str_array->strings = malloc(str_array->count * sizeof(char *));
 
 	for (i = 0; i < RARRAY_LEN(rb_array); ++i) {
 		VALUE rb_string = rb_ary_entry(rb_array, i);
--- a/ext/rugged/rugged.c
+++ b/ext/rugged/rugged.c
@@ -436,9 +436,11 @@ void rugged_rb_ary_to_strarray(VALUE rb_array, git_strarray *str_array)
 		return;
 
 	if (TYPE(rb_array) == T_STRING) {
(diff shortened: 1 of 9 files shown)


Ekene1998 (approved):

djberg96:
Anyone home?

carlosmn:
Are you sure calling `xfree` on `NULL` can cause an "invalid pointer" error? Doing nothing when passed `NULL` is the behaviour that any allocator should have, and indeed chasing down `xfree` in ruby through `ruby_xfree` to `ruby_sized_xfree`, we can see in https://github.com/ruby/ruby/blob/373eacb8416f8b20e89e2b52dc007c5958a79bdb/gc.c#L5442 that if does nothing if this is `NULL`.

Now, if we end up mixing up something like `Qnil` with `NULL` and then try to free that, that I can see causing trouble, but this would not fix that as this is doing the same check that `xfree` ends up doing, just a few calls higher in the stack.

carlosmn:
What I'm noticing now is that we are calling the wrong free function completely, and causing memory leaks. Rugged should always be using the ruby allocator but if we don't end up doing that, `xfree` is the wrong allocator to use.

On top of that, we should be using `git_strarray_dispose` in order to actually free the memory and not just he array containing them.

djberg96:
Indeed, my original PR was superficial. I'm working on updating it...

djberg96:
@carlosmn ok, updated (courtesy of Raptor Mini).

Ran the specs a couple times, all passed locally. I haven't run it through valgrind yet (won't install on latest Ubuntu, will have to wait until I'm home).

djberg96:
I added a stress test.

carlosmn:
I don't know who Raptor Mini is but it turns out I misremembered and we should not use `git_strarray_dispose`. This PR also moves away from the ruby allocator to the system one which is incorrect as we cannot guarantee that the libgit2 allocator is the same as the system one. I'm surprised this does work without crashing.

We are re-using the string pointers from ruby which is why we don't own the memory of the strings and why we were freeing the one pointer we do own manually.

That is confusing so I've opened #999 to explain and make the relationship explicit.

Either way, none of this would explain why `xfree` would cause any crashes here when freeing a pointer we allocated.

djberg96:
ok, as long as it's fixed :)

PR #1000, opened 2026-04-29
ext: hide statically-linked symbols to prevent llhttp collision

stanhu:
libgit2 >= 1.8 bundles llhttp as its default HTTP parser. Without visibility controls, all 119 llhttp_* symbols are exported from rugged.so and collide with the llhttp-ffi gem (used by the http gem) when both are loaded in the same process.

Fix by:
- Passing -DCMAKE_C_VISIBILITY_PRESET=hidden to cmake so libgit2 and its bundled deps (including llhttp) are compiled with -fvisibility=hidden on all platforms.
- Appending -Wl,--exclude-libs,ALL to $LDFLAGS on Linux as a belt-and-suspenders linker-level guard (macOS ld does not support this flag, but the cmake flag covers it there).

### DIFF
--- a/ext/rugged/extconf.rb
+++ b/ext/rugged/extconf.rb
@@ -110,7 +110,7 @@ else
     Dir.chdir("build") do
       # On Windows, Ruby-DevKit is MSYS-based, so ensure to use MSYS Makefiles.
       generator = "-G \"MSYS Makefiles\"" if Gem.win_platform?
-      run_cmake(5 * 60, ".. -DBUILD_TESTS=OFF -DUSE_THREADS=ON -DBUILD_SHARED_LIBS=OFF -DCMAKE_C_FLAGS=-fPIC -DCMAKE_BUILD_TYPE=RelWithDebInfo #{cmake_flags.join(' ')} #{generator}")
+      run_cmake(5 * 60, ".. -DBUILD_TESTS=OFF -DUSE_THREADS=ON -DBUILD_SHARED_LIBS=OFF -DCMAKE_C_FLAGS=-fPIC -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_VISIBILITY_PRESET=hidden #{cmake_flags.join(' ')} #{generator}")
       sys(MAKE)
 
       # "normal" libraries (and libgit2 builds) get all these when they build but we're doing it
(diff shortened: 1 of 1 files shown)


stanhu:
### Before

```
$ objdump -T tmp/x86_64-linux/stage/lib/rugged/rugged.so | grep llhttp | head
000000000010bb30 g    DF .text  0000000000000010  Base        llhttp__internal__c_test_flags
000000000010b8e0 g    DF .text  000000000000014b  Base        llhttp__after_headers_complete
000000000010bc90 g    DF .text  0000000000000009  Base        llhttp__internal__c_load_header_state
000000000010b400 g    DF .text  000000000000004e  Base        llhttp__on_version
000000000010b170 g    DF .text  000000000000001a  Base        llhttp_set_lenient_transfer_encoding
000000000010b7c0 g    DF .text  0000000000000072  Base        llhttp__debug
000000000010b630 g    DF .text  0000000000000023  Base        llhttp__on_chunk_header
000000000010bef0 g    DF .text  000000000000000b  Base        llhttp__internal__c_update_type
000000000010a660 g    DF .text  0000000000000009  Base        llhttp_get_http_major
000000000010aac0 g    DF .text  0000000000000308  Base        llhttp_method_name
```

After:

```
$ objdump -T tmp/x86_64-linux/stage/lib/rugged/rugged.so | grep llhttp | head
$
```

stanhu:
@carlosmn Could you look at this?

stanhu:
@ethomson We ran into a significant regression after updating to Rugged v1.9 when the `http` gen were used, which uses `llhttp-ffi`:

`rugged.so` globally exports 119 `llhttp_*` symbols from libgit2's statically-linked bundled llhttp — a symbol visibility failure caused by the absence of `-fvisibility=hidden` or a linker version script at build time. When `llhttp-ffi` (loaded transitively by the `http` gem) is mapped into the same process, the dynamic linker's interposition rules cause its `llhttp_*` calls to resolve to rugged's already-resident symbols rather than its own. Because the two copies are different major versions, the `llhttp_settings_t` callback signatures and `llhttp_t` state struct layout don't match — so when `llhttp-ffi` registers its span callbacks and invokes the parser, the wrong function pointers are called against a misinterpreted state struct. llhttp surfaces this as `HPE_USER` and halts parsing, causing every download to fail:

```
Error downloading file: error reading from socket:
Error Parsing data: HPE_USER Span callback error in on_header_field
```

ethomson:
That sounds like a problem, indeed. Sounds like something that libgit2 should be addressing. I won't be able to take a look for a few days, unfortunately.

brodock:
@ethomson any change this could be reviewed?

carlosmn:
It does seem like something that libgit2 should be doing itself. If you build the shared object, you don't get these symbols exported, so I went looking and we might just want to move the line at https://github.com/libgit2/libgit2/blob/44c05e5d12f2b8b86b9730bb50f27daf74143782/cmake/DefaultCFlags.cmake#L116 to be outside of that IF. Testing locally it seems to do the trick.

It looks like `-fvisibility=hidden` was put as SO-only accidentally given that https://github.com/libgit2/libgit2/commit/b41e24a65c73c4ce052b02134ca559803400eecd only talks about `-fPIC`.

PR #1003, opened 2026-06-01
Use TypedData APIs exclusively

### LINKED ISSUE #1002: Doesn't build on Ruby 4.1 because untyped data API got removed
I'm testing against ruby-head and rugged doesn't build against Ruby 4.1 anymore. Ruby removed some deprecated functions recently (https://github.com/ruby/ruby/pull/15447), which caused this.

I found https://github.com/libgit2/rugged/pull/998 which already moved most things over but it looks like the comment about the backend has now become relevant (: There's also one gem-controlled instance here that needs to be changed https://github.com/libgit2/rugged/blob/0772a8e417d0db497ea32e3c6e9133970335035d/ext/rugged/rugged.c#L690

jhawthorn:
https://github.com/libgit2/rugged/pull/998 updated most of the APIs to use TypedData 🙌. However a few of the legacy uses were still left. This PR converts the remaining two cases.

The legacy API has been removed from Ruby HEAD (likely to be released with Ruby 4.1 in December).

One thing to note is that I don't believe `rb_mShutdownHook` worked before this change. When the wrapped struct is NULL the dfree function is skipped.

A new release after this is merged would be very appreciated for Ruby HEAD compatibility 🙏

Fixes https://github.com/libgit2/rugged/issues/1002

### DIFF
--- a/ext/rugged/rugged.c
+++ b/ext/rugged/rugged.c
@@ -336,6 +336,16 @@ static void cleanup_cb(void *unused)
 	git_libgit2_shutdown();
 }
 
+static const rb_data_type_t rugged_shutdown_hook_type = {
+	.wrap_struct_name = "Rugged::ShutdownHook",
+	.function = {
+		.dmark = NULL,
+		.dfree = cleanup_cb,
+		.dsize = NULL,
+	},
+	.flags = RUBY_TYPED_FREE_IMMEDIATELY,
+};
+
 void rugged_exception_raise(void)
 {
 	VALUE err_klass, err_obj;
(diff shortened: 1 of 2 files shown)


nobu:
I think the clean-up can be done in the end-proc (see #1004).

carlosmn:
Thanks, I had left the backends code alone because I had no idea how you're even supposed to wrap that kind of thing, and I guess the answer is you just don't.

byroot:
> A new release after this is merged would be very appreciated for Ruby HEAD compatibility

I'd like to second this request, because it seems `rugged` cannot be used as a git gem in the meantime :/

byroot:
nevermind, it seems like I had a corrupted cache. A first attempt a `gem 'rugged', github: 'libgit2/rugged'` created `vendor/libgit2/build`, which then broke the later attempt using `gem 'rugged', github: 'libgit2/rugged', submodules: true`, so I'm unblocked for now.

A release definitely would be nice though.

ioquatix:
+1 for @nobu's PR and a patch release :)

